PR #39, opened 2020-06-14
Assignment_1 completed

RajAJun20:
-> Added main method
-> Print name


src/rajAJun20/Test.java
```diff
@@ -0,0 +1,11 @@
+package rajAJun20;
+
+public class Test {
+	 
+	public static void main(String[] args) {
+
```
KrishnaKTechnocredits, line 6: Remove space.

PR #41, opened 2020-06-14
Assignment 2 for Git practice.

NikhilMJun20:


src/nikhilMJun20/GitTest.java
```diff
@@ -0,0 +1,8 @@
+package nikhilMJun20;
+
+public class GitTest {
+
```
KrishnaKTechnocredits: Remove space

NikhilMJun20: -> Suggested changes done.

PR #43, opened 2020-06-14
Assignment1completed

AbhishekSJune2020:
-->added main method.
-->print my name


src/abhishekSJun20/Test.java
```diff
@@ -0,0 +1,8 @@
+package abhishekSJun20;
+
+public class Test {
+
```
KrishnaKTechnocredits, line 4: Remove space.

PR #48, opened 2020-06-14
Practice

BarkhaPJun20:
Sent code for review (Git Practice)


src/barkhaPJun20/Test.java
```diff
@@ -0,0 +1,10 @@
+package barkhaPJun20;
+
+public class Test {
+	
+	public static void main(String[] args) {
+		System.out.println("Barkha");
+	}
+
```
KrishnaKTechnocredits: Remove space.

BarkhaPJun20: done

PR #53, opened 2020-06-15
Adding Git Test 2

AbhijitMJun20:
Adding git test 2


src/abhijitMJun20/Test2.java
```diff
@@ -0,0 +1,10 @@
+package abhijitMJun20;
+
+public class Test2 {
+
```
KrishnaKTechnocredits: Remove space.

AbhijitMJun20: Done

PR #61, opened 2020-06-15
new file

DeavinaSJun20:


src/deavinaSJun20/TestExample.java
```diff
@@ -0,0 +1,14 @@
+package deavinaSJun20;
+
+public class TestExample {
+
+	void Display() {
+
```
KrishnaKTechnocredits: Remove space.

DeavinaSJun20: updated as per comment

PR #65, opened 2020-06-15
Demo Assignment to print number

HarshadBJun20:
-> Demo class printing a number

### DIFF
--- a/src/harshadBJun20/Demo.java
+++ b/src/harshadBJun20/Demo.java
@@ -1,0 +1,15 @@
+package harshadBJun20;
+
+public class Demo {
+	
+	int Number=10;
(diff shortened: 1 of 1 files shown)


src/harshadBJun20/Demo.java
```diff
@@ -0,0 +1,15 @@
+package harshadBJun20;
+
+public class Demo {
+	
+	int Number=10;
+
```
KrishnaKTechnocredits: Remove space.

HarshadBJun20: Done

KrishnaKTechnocredits (approved):

PR #66, opened 2020-06-15
Viresh assignement 1

VireshJJun20:
Kindly review

### DIFF
--- a/src/vireshJJun20/Test.java
+++ b/src/vireshJJun20/Test.java
@@ -2,12 +2,11 @@ package vireshJJun20;
 
 public class Test {
 	
-	void Display() {
+	void display() {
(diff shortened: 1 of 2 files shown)


src/vireshJJun20/Test.java
```diff
@@ -0,0 +1,14 @@
+package vireshJJun20;
+
+public class Test {
+	
+	void Display() {
+		System.out.println("Viresh");
+	}
+	public static void main(String[] args) {
+		Test test= new Test();
+
```
KrishnaKTechnocredits: Remove space.

VireshJJun20: Removed space as suggested.

PR #70, opened 2020-06-15
Assignment_2_Swapna

SwapnaWJun20:
New file - To check the steps

### DIFF
--- a/src/swapnaWJun20/Demo.java
+++ b/src/swapnaWJun20/Demo.java
@@ -2,11 +2,11 @@ package swapnaWJun20;
 
 public class Demo {
 	
-	void Display(){
+	void display(){
 		System.out.println(" Test ");
 	}
 	public static void main(String[] args) {
 		Demo demo = new Demo();
-		demo.Display();
+		demo.display();
 	}
 }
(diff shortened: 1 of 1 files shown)


src/swapnaWJun20/Demo.java
```diff
@@ -0,0 +1,12 @@
+package swapnaWJun20;
+
+public class Demo {
+
```
KrishnaKTechnocredits, line 4: Remove space.

src/swapnaWJun20/Demo.java
```diff
@@ -0,0 +1,12 @@
+package swapnaWJun20;
+
+public class Demo {
+	
+	void Display(){
```
KrishnaKTechnocredits: Method name should be start with small letter.

src/swapnaWJun20/Demo.java
SwapnaWJun20: Done

src/swapnaWJun20/Demo.java
SwapnaWJun20: Done

PR #72, opened 2020-06-15
Test Data file created on git and commit action performed

VaibhavTati:
Please check


src/vaibhavTJun20/Test.java
```diff
@@ -0,0 +1,5 @@
+package vaibhavTJun20;
+
+public class Test {
+
```
KrishnaKTechnocredits, line 4: Add main() method.

PR #79, opened 2020-06-15
Added one file

KrishnaKTechnocredits:


src/technoCredits/Test3.java
```diff
@@ -0,0 +1,12 @@
+package technoCredits;
+
+public class Test3 {
+
+	void Display() {
```
KrishnaKTechnocredits: Method name should be start with small letter.

KrishnaKTechnocredits: Done.

PR #100, opened 2020-06-16
Harshad_Assignment_8_BankingSystem_Done

HarshadBJun20:
->Banking System code is completed
->Addition to given requirement have taken 2 users and separate balance amount for each user.
->User's Individual transaction summary and Overall transaction summary is being displayed.

### DIFF
--- a/src/harshadBJun20/BankingSystem.java
+++ b/src/harshadBJun20/BankingSystem.java
@@ -19,7 +19,7 @@ All transaction summary : Credit - 7 times, Debit - 3 times, printBalance - 1 ti
 
 package harshadBJun20;
 
-class BankingSystem{
+public class BankingSystem{
 	
 	int debitIndividualCount,creditIndividualCount,balanceIndividualCount, balanceAmt;
 	static int debitTotalCount=0, creditTotalCount=0, balanceTotalCount=0;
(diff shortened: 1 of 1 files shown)


src/harshadBJun20/BankingSystem.java
```diff
@@ -0,0 +1,105 @@
...
+		System.out.println("Current Balance of user "+userId+" is: "+balanceAmt);
+		System.out.println();
+		balanceIndividualCount++;
+		balanceTotalCount++;
+	}
+	
+	//Method to Print individual transaction summary
+	void individualTransactionSummary(){
+		
+		if(userId == 1){
```
KrishnaKTechnocredits: Here we don't need if blocks for users because you already created userId as instance variable. 
And if you want to print individualTransactionSummary() user vise then you should right print a message like 
System.out.println(userId + "transaction summary : " +"Credit: " +creditIndividualCount+" times, " +"Debit: " + debitIndividualCount+" times," +" Print Balance: "+balanceIndividualCount+" times");

src/harshadBJun20/BankingSystem.java
```diff
@@ -0,0 +1,105 @@
...
+
+	//Method to set UserId and balanceAmount.
+	void setData(int uId, int balance){
+		userId = uId;
+		balanceAmt = balance;
+	}
+	
+	//Method to Debit the Amount from Balance
+	void debitAmount(int debitAmt){
+		this.balanceAmt= this.balanceAmt - debitAmt;
```
KrishnaKTechnocredits: You should add a condition for insufficient balance.

src/harshadBJun20/BankingSystem.java
HarshadBJun20: -> Removed if conditions and printed the userId for individualTransactionSummary() user wise.
-> Done. Thanks

src/harshadBJun20/BankingSystem.java
HarshadBJun20: ->Done

KrishnaKTechnocredits (commented):
Code looks good to me.

PR #101, opened 2020-06-16
Pranita_Assignment_8_Done

PranitaPJun20:
In this assignment added Print Current balance logic.Used balance as a static variable.
Added 5 methods as per requirement and shown two users transaction summary.

### DIFF
--- a/src/pranitaPJun20/BankingSystem.java
+++ b/src/pranitaPJun20/BankingSystem.java
@@ -1,74 +1,85 @@
 package pranitaPJun20;
 
 public class BankingSystem {
-	//Instance Variables
-	int debitCount=0;
-	int creditCount=0;
-	int printBalanceCount=0;
-	//Static variables
-	static int balance=100000;
-	static int staticDebitCount=0;
-	static int staticCreditCount=0;
-	static int staticPrintBalCount=0;
+//Instance Variables
+	int debitCount;
+	int creditCount;
+	int printBalanceCount;
+	int balance;
+//Static variables
+	static int staticDebitCount;
+	static int staticCreditCount;
+	static int staticPrintBalCount;
 	
-		//Method to Debit amount
+//Method to Debit amount
 		void debitAmnt(int amountDebited ){
 			debitCount= debitCount+1;
 			//debitCount++ ; //we can use this also
 			//debitCount+=1 ;//we can use this command also	
 			staticDebitCount=staticDebitCount+1;
 			balance= balance-amountDebited;
-		}
-		//Method to credit amount
+				if(balance<=100) {
+					System.out.println("Account has insufficient balance");			
+				}
+			}
+		
+//Method to credit amount
 		void creditAmnt(int amountCredited){
 			creditCount= creditCount+1;
 			staticCreditCount=staticCreditCount+1;
-			balance= balance-amountCredited;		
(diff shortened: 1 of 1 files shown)


src/pranitaPJun20/BankingSystem.java
```diff
@@ -0,0 +1,76 @@
...
+public class BankingSystem {
+	//Instance Variables
+	int debitCount=0;
+	int creditCount=0;
+	int printBalanceCount=0;
+	//Static variables
+	static int balance=100000;
+	static int staticDebitCount=0;
+	static int staticCreditCount=0;
+	static int staticPrintBalCount=0;
```
KrishnaKTechnocredits: Remove initialization of all instance variables. And fro initial balance you can call credit() method first.

src/pranitaPJun20/BankingSystem.java
```diff
@@ -0,0 +1,76 @@
...
+	
+		//Method to Debit amount
+		void debitAmnt(int amountDebited ){
+			debitCount= debitCount+1;
+			//debitCount++ ; //we can use this also
+			//debitCount+=1 ;//we can use this command also	
+			staticDebitCount=staticDebitCount+1;
+			balance= balance-amountDebited;
```
KrishnaKTechnocredits, line 20: You should write the condition to verify the insufficient balance.

src/pranitaPJun20/BankingSystem.java
```diff
@@ -0,0 +1,76 @@
...
+		}
+		//Method to credit amount
+		void creditAmnt(int amountCredited){
+			creditCount= creditCount+1;
+			staticCreditCount=staticCreditCount+1;
+			balance= balance-amountCredited;
```
KrishnaKTechnocredits: Do copy-paste code. Here it should be addition operation.

src/pranitaPJun20/BankingSystem.java
PranitaPJun20: Done.Added condition to handle insufficient balance in account

src/pranitaPJun20/BankingSystem.java
PranitaPJun20: Done

src/pranitaPJun20/BankingSystem.java
PranitaPJun20: Done

src/pranitaPJun20/BankingSystem.java
PranitaPJun20: Yes.Changed

src/pranitaPJun20/BankingSystem.java
```diff
@@ -0,0 +1,87 @@
...
+	static int staticPrintBalCount;
+	
+//Method to Debit amount
+		void debitAmnt(int amountDebited ){
+			debitCount= debitCount+1;
+			//debitCount++ ; //we can use this also
+			//debitCount+=1 ;//we can use this command also	
+			staticDebitCount=staticDebitCount+1;
+			balance= balance-amountDebited;
+				if(balance<=100) {
```
KrishnaKTechnocredits, line 21: This is not the condition that we expected. Please review https://github.com/KrishnaKTechnocredits/JAVATechnoJun20/pull/100

PR #104, opened 2020-06-16
Suparna_assignment_8

SuparnaNikam:
Suparna_assignment_8_Done

### DIFF
--- a/src/suparnaNJune20/Bank.java
+++ b/src/suparnaNJune20/Bank.java
@@ -1,0 +1,72 @@
+package suparnaNJune20;
+
+public class Bank {
+
+	static double atmBal=100000;
+	static int debitCnt, creditCnt, printBillCnt;
+	int debit,credit, printBill;
+	
+	void debitTransaction(int amt, String name) {
+		if (amt>atmBal)	{
+			System.out.println("No Sufficient BAl ");
+			return;
+		}
+		else {
+			atmBal-=amt;
+			debit++;
+			debitCnt++;
+			System.out.println(amt+ " Rs debited sucesfully by " + name);
+		}
+	}
+		void  creditTransaction (int amt, String name) {
+			 atmBal+=amt;
+				System.out.println(amt+ " Rs credited sucesfully by"+ name);
+				credit++;
+				creditCnt++;
+	}		
+		void  printBillTransaction() {
+			System.out.println(" Current Balance in Atm :"+atmBal);
+			printBill++;
+			printBillCnt++;
+		}
+	
+	void individualTransactionSumary(String custName) {
+		System.out.println("\n\n Total No of  trnasaction done by  "+custName +" are "+ (debit+credit+ printBill));
+		System.out.println("Total No of Credit transaction " + credit);
+		System.out.println("Total No of Debit transaction " + debit);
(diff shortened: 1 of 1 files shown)


src/suparnaNJune20/Bank.java
```diff
@@ -0,0 +1,72 @@
+package suparnaNJune20;
+
+public class Bank {
+
+	static double atmBal=100000;
```
KrishnaKTechnocredits: Do not hard code amount balance, you should write setdata() method for this.

src/suparnaNJune20/Bank.java
```diff
@@ -0,0 +1,72 @@
...
+		System.out.println("Total No of Credit transaction " + creditCnt);
+		System.out.println("Total No of Debit transaction " + debitCnt);
+		System.out.println("Total No of PrintBill transaction "+ printBillCnt);
+	}
+	public static void main(String[] args) {
+		// TODO Auto-generated method stub
+		Bank customer1 = new Bank();
+		Bank customer2 = new Bank();
+		System.out.println("Current ATM bal : "+customer1.atmBal );
+		for (int i = 40; i<45; i++)
```
KrishnaKTechnocredits: When we write for a look range between 40 to 45, it will be a confusing part for the end-user that why you have selected this range. So here, for loop is not needed.

src/suparnaNJune20/Bank.java
```diff
@@ -0,0 +1,72 @@
...
+		else {
+			atmBal-=amt;
+			debit++;
+			debitCnt++;
+			System.out.println(amt+ " Rs debited sucesfully by " + name);
+		}
+	}
+		void  creditTransaction (int amt, String name) {
+			 atmBal+=amt;
+				System.out.println(amt+ " Rs credited sucesfully by"+ name);
```
KrishnaKTechnocredits: Work on indentation.

src/suparnaNJune20/Bank.java
```diff
@@ -0,0 +1,72 @@
...
+				credit++;
+				creditCnt++;
+	}		
+		void  printBillTransaction() {
+			System.out.println(" Current Balance in Atm :"+atmBal);
+			printBill++;
+			printBillCnt++;
+		}
+	
+	void individualTransactionSumary(String custName) {
```
KrishnaKTechnocredits: Here this method not required custName parameter. If you want to print customer name then you should take as instance variable and should have setName() method for this.

src/suparnaNJune20/Bank.java
SuparnaNikam: ddone

src/suparnaNJune20/Bank.java
SuparnaNikam: done

src/suparnaNJune20/Bank.java
SuparnaNikam: done

src/suparnaNJune20/Bank.java
SuparnaNikam: done

KrishnaKTechnocredits (commented):
Code looks good to me.

PR #106, opened 2020-06-16
Sweety_Assignment_8_Done

SweetyMJun20:
Assignment 8

### DIFF
--- a/src/sweetyMJun20/BankingSystem.java
+++ b/src/sweetyMJun20/BankingSystem.java
@@ -1,0 +1,52 @@
+package sweetyMJun20;
+
+public class BankingSystem {
+	int creditcnt;
+	int debitcnt;
+	int printcnt;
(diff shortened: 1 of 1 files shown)


src/sweetyMJun20/BankingSystem.java
```diff
@@ -0,0 +1,52 @@
...
+	static int userdebitcnt;
+	static int userprintcnt;
+	
+	void debitAcc(int debtamt){
+        debitcnt++;
+		userdebitcnt++;
+	}
+	void creditAcc(int creditamt){
+		creditcnt++;
+		usercreditcnt++;
```
KrishnaKTechnocredits: You forgot to perform an operation in debitAcc() and creditAcc() METHODS. You should have a balance instance variable.  You should write the condition to verify the insufficient balance.

PR #107, opened 2020-06-16
Vaishnavi assignment 8

VaishnaviVJun20:
-Vaishnavi_Assignment_8_Done

### DIFF
--- a/src/vaishnaviVJun20/Bank.java
+++ b/src/vaishnaviVJun20/Bank.java
@@ -1,7 +1,6 @@
 package vaishnaviVJun20;
 
-public class Bank {
-    
+public class Bank {   
 	int creditCount = 0;
 	int debitCount = 0;
 	int printBalanceCount = 0;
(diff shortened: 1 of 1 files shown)


src/vaishnaviVJun20/Bank.java
```diff
@@ -0,0 +1,64 @@
+package vaishnaviVJun20;
+
+public class Bank {   
+	int creditCount = 0;
+	int debitCount = 0;
+	int printBalanceCount = 0;
+	static int creditCount1 = 0;
+	static int debitCount1 = 0;
+	static int printBalanceCount1 = 0;
```
KrishnaKTechnocredits: Remove the initialization of all instance variables.

src/vaishnaviVJun20/Bank.java
```diff
@@ -0,0 +1,64 @@
...
+	int balance = 50000;
```
KrishnaKTechnocredits: No Need to initialize this variable. perform credit() method first.

src/vaishnaviVJun20/Bank.java
VaishnaviVJun20: Done

src/vaishnaviVJun20/Bank.java
VaishnaviVJun20: Done

KrishnaKTechnocredits (commented):
///

PR #108, opened 2020-06-16
Abhijit_Assignment_8_Done

AbhijitMJun20:
Assignment_8_Done

### DIFF
--- a/src/abhijitMJun20/Bankingsystem.java
+++ b/src/abhijitMJun20/Bankingsystem.java
@@ -1,14 +1,16 @@
 package abhijitMJun20;
 
 public class Bankingsystem {
-	int accountbalance=1000;
-	static int credittransactionsummary=0,debittransactionsummary=0,printbalancesummary=0;
-	int creditcount=0,debitcount=0,printBalancecount=0;
+	int accountbalance;
+	static int credittransactionsummary,debittransactionsummary,printbalancesummary;
+	int creditcount,debitcount,printBalancecount;
 	
 	void debitAmount(int debitamount) { 
 		accountbalance=accountbalance-debitamount;
+		if(accountbalance != 0){
 		debitcount++;
 		debittransactionsummary++;
+		}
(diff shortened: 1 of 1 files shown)


src/abhijitMJun20/Bankingsystem.java
```diff
@@ -0,0 +1,52 @@
+package abhijitMJun20;
+
+public class Bankingsystem {
+	int accountbalance=1000;
```
KrishnaKTechnocredits: No Need to initialize this variable. perform credit() method first.

src/abhijitMJun20/Bankingsystem.java
```diff
@@ -0,0 +1,52 @@
...
+	static int credittransactionsummary=0,debittransactionsummary=0,printbalancesummary=0;
+	int creditcount=0,debitcount=0,printBalancecount=0;
```
KrishnaKTechnocredits: Remove the initialization of all instance variables.

src/abhijitMJun20/Bankingsystem.java
```diff
@@ -0,0 +1,52 @@
...
+	
+	void debitAmount(int debitamount) { 
+		accountbalance=accountbalance-debitamount;
```
KrishnaKTechnocredits, line 9: You should write the condition to verify the insufficient balance.

src/abhijitMJun20/Bankingsystem.java
AbhijitMJun20: Done

src/abhijitMJun20/Bankingsystem.java
AbhijitMJun20: Done

src/abhijitMJun20/Bankingsystem.java
AbhijitMJun20: Done

PR #114, opened 2020-06-16
Assignment_8_MeghaDone

MeghaJJune20:
Assignment Bank Completed

### DIFF
--- a/src/meghaJJune20/Bank.java
+++ b/src/meghaJJune20/Bank.java
@@ -7,9 +7,9 @@ public class Bank {
 	int countbalance;
 	int userid;
 	
-	static int debittotal=0;
-	static int credittotal=0;
-	static int baltotal=0;
+	static int debittotal;
+	static int credittotal;
+	static int baltotal;
 		
 	void setData( int userid1,int balance)
 	{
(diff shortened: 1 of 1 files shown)


src/meghaJJune20/Bank.java
```diff
@@ -0,0 +1,77 @@
...
+public class Bank {
+	
+	int countdebit;
+	int countcredit;
+	int countbalance;
+	int userid;
+	
+	static int debittotal=0;
+	static int credittotal=0;
+	static int baltotal=0;
```
KrishnaKTechnocredits: Remove the initialization of all instance variables.

src/meghaJJune20/Bank.java
```diff
@@ -0,0 +1,77 @@
...
+	{
+		System.out.println();
+		System.out.println("Balance of User " +userid+" is "+countbalance);
+		countbalance++;
+		baltotal++;
+	}
+	void individualTransactionSummary()
+	{
+		System.out.println();
+		if(userid==1)
```
KrishnaKTechnocredits: No Need to write if block here. Because you already created instance variable for the userID.

src/meghaJJune20/Bank.java
MeghaJJune20: Done as per comments

src/meghaJJune20/Bank.java
MeghaJJune20: Done as per comments

PR #117, opened 2020-06-16
Yash_Assignment08_Done

YashCJun20:
Assignment Completed
Please review

### DIFF
--- a/src/yashCJun20/BankingSystem.java
+++ b/src/yashCJun20/BankingSystem.java
@@ -23,18 +23,28 @@ class BankingSystem{
 	
 	int indvDebit;
 	int indvCredit;
-	int balance =5000;
+	int balance;
 	int indvBalance;
 	static int debit;
 	static int credit;
-	static int finalbalance;
+	static int finalBalance;
 	
 	
 	void debitAmount(int debtAmt){
 		
-		balance = balance-debtAmt;
-		indvDebit=indvDebit+1;
-		debit=debit+1;
+		
+		if (debtAmt <=balance) {
+			
+			balance = balance-debtAmt;
+			indvDebit=indvDebit+1;
+			debit=debit+1;
+			
+	}
+		
+		else
+			System.out.println("The requested amount cannot be withdraw due to insufficent balance");
+		
+			
 	}
 	
 	void creditAmount(int crdtAmt){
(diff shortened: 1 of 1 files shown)


src/yashCJun20/BankingSystem.java
```diff
@@ -0,0 +1,97 @@
...
+5) allTransactionSummary() method should show by both users total how many times debit, credit & printBalance method get called.
+Hint : Create 3 static variable will be required.
+Output:
+All transaction summary : Credit - 7 times, Debit - 3 times, printBalance - 1 time */
+
+class BankingSystem{
+	
+	int indvDebit;
+	int indvCredit;
+	int balance =5000;
```
KrishnaKTechnocredits: No Need to initialize this variable. perform credit() method first.

src/yashCJun20/BankingSystem.java
```diff
@@ -0,0 +1,97 @@
...
+	int indvBalance;
+	static int debit;
+	static int credit;
+	static int finalbalance;
+	
+	
+	void debitAmount(int debtAmt){
+		
+		balance = balance-debtAmt;
```
KrishnaKTechnocredits: You should write the condition to verify the insufficient balance.

src/yashCJun20/BankingSystem.java
```diff
@@ -0,0 +1,97 @@
...
+	void printBalance(){
+		
+		System.out.println("Current Balance " +": " +balance);
+		indvBalance=indvBalance+1;
+		finalbalance=finalbalance+1;
+	}
+	
+	void individualTransactionSummary(String user1){
+		
+		if(user1.equals("Yash"))
```
KrishnaKTechnocredits: You don't need if block here because you already set an instance variables.

src/yashCJun20/BankingSystem.java
YashCJun20: Done

src/yashCJun20/BankingSystem.java
YashCJun20: Done

src/yashCJun20/BankingSystem.java
YashCJun20: Done

KrishnaKTechnocredits (commented):
Code looks good to me.

PR #125, opened 2020-06-16
Piyush_Assignment_8_Done

PiyushPJun20:
Piyush_Assignment_8_Done

### DIFF
--- a/src/BankingSystem.java
+++ b/src/BankingSystem.java
@@ -1,28 +1,41 @@
 
 public class BankingSystem {
 
-	int creditCount, debitCount, printCount;
-	static int userCrdtCount;
-	static int userDebtCount;
-	static int userprintCount;
+	int creditCount, debitCount, printCount, balance;
+	static int userCrdtCount, userDebtCount, userprintCount, countUser;
+	
+	void setInitialBalance(int amount){
+		countUser++;
+		System.out.println("");
+		System.out.println(" Initial Balance of User "+countUser+" is: "+amount);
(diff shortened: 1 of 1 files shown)


src/BankingSystem.java
```diff
@@ -0,0 +1,55 @@
...
+	static int userDebtCount;
+	static int userprintCount;
+
+	void debitAccount(int debitAmt){
+        debitCount++;
+		userDebtCount++;
+	}
+	void creditAccount(int creditAmt){
+		creditCount++;
+		userCrdtCount++;
```
KrishnaKTechnocredits, line 27: You forgot to perform an operation in debitAcc() and credit() METHODS. You should have a balance instance variable. You should write the condition to verify the insufficient balance.

src/BankingSystem.java
```diff
@@ -0,0 +1,55 @@
...
+	}
+	void printBalance(int crAmt,int debtAmt){
+	    System.out.println(""); 
+	    System.out.println(" Current Balance is : " + ((crAmt-debtAmt)));
```
KrishnaKTechnocredits: This line not required here.

PiyushPJun20: Changes Incorporated.

src/BankingSystem.java
PiyushPJun20: Changes Incorporated.

KrishnaKTechnocredits (commented):
Code looks good to me.

PR #129, opened 2020-06-16
Kartikey_Assignment_8_Done

KartikeyDJune20:


src/kartikeyDJune20/BankingSystem.java
```diff
@@ -0,0 +1,54 @@
+package kartikeyDJune20;
+
+public class BankingSystem {
+	int totalBalance = 9000;
```
KrishnaKTechnocredits, line 4: No Need to initialize this variable. perform credit() method first.

PR #134, opened 2020-06-17
Updated code per review comments

VireshJJun20:
Updated code per review comments, removed the old program as well

### DIFF
--- a/src/vireshJJun20/BankOps.java
+++ b/src/vireshJJun20/BankOps.java
@@ -1,0 +1,81 @@
+package vireshJJun20;
+
+public class BankOps {
+	static int availableBal=10000;
+	static int allCredit=0;
+	static int allDebit=0;
+	static int allPrintBal=0;
+	int creditCnt;
+	int	debitCnt;
+	int printCnt;
+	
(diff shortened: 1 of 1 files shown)


src/vireshJJun20/BankOps.java
```diff
@@ -0,0 +1,81 @@
...
+	static int allCredit=0;
+	static int allDebit=0;
+	static int allPrintBal=0;
+	int creditCnt;
+	int	debitCnt;
+	int printCnt;
+	
+	int creditTr(int cr) {
+		System.out.println(" - User willing to credit amount of " + cr);
+		if(availableBal+cr<10000) {
```
KrishnaKTechnocredits, line 14: Are you really think this logic is needed?

src/vireshJJun20/BankOps.java
```diff
@@ -0,0 +1,81 @@
+package vireshJJun20;
+
+public class BankOps {
+	static int availableBal=10000;
```
KrishnaKTechnocredits, line 4: No Need to initialize this variable. perform credit() method first.

PR #136, opened 2020-06-17
Vaishnavi_Assignment_9_Done

VaishnaviVJun20:
-Please review

### DIFF
--- a/src/vaishnaviVJun20/Math.java
+++ b/src/vaishnaviVJun20/Math.java
@@ -1,13 +1,11 @@
 package vaishnaviVJun20;
 
 public class Math {
-    double sum(double variable1, double variable2, double variable3) {
-    	double sum = variable1+variable2+variable3;
-    	return sum;
+    double sum(double number1, double number2, double number3) {
+    	return number1+number2+number3;
     }
-    double average(double variable1, double variable2, double variable3) {
-    	System.out.println("Sum of three variable is: " +sum(variable1,variable2,variable3));
-    	double average = (variable1+variable2+variable3)/3;
+    double average(double number1, double number2, double number3) {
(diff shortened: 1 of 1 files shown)


src/vaishnaviVJun20/Math.java
```diff
@@ -0,0 +1,40 @@
+package vaishnaviVJun20;
+
+public class Math {
+    double sum(double variable1, double variable2, double variable3) {
+    	double sum = variable1+variable2+variable3;
+    	return sum;
+    }
+    double average(double variable1, double variable2, double variable3) {
+    	System.out.println("Sum of three variable is: " +sum(variable1,variable2,variable3));
+    	double average = (variable1+variable2+variable3)/3;
```
KrishnaKTechnocredits: you should say average = sum(variable1,variable2,variable3)/3; Otherwise there is no need of sum method. we asked you call sum method from average. Please correct it

src/vaishnaviVJun20/Math.java
```diff
@@ -0,0 +1,40 @@
+package vaishnaviVJun20;
+
+public class Math {
+    double sum(double variable1, double variable2, double variable3) {
```
KrishnaKTechnocredits: It looks better if we use number1 instead of variable1

src/vaishnaviVJun20/Math.java
VaishnaviVJun20: Done

src/vaishnaviVJun20/Math.java
VaishnaviVJun20: Done

KrishnaKTechnocredits (commented):
The code looks good to me now.

PR #137, opened 2020-06-17
Viresh_Assignment_9_Done

VireshJJun20:
Please review and comment.

### DIFF
--- a/src/vireshJJun20/ArithOps.java
+++ b/src/vireshJJun20/ArithOps.java
@@ -10,9 +10,9 @@ class ArithOps{
 	}
 	
 	double average(double a, double b, double c){
-		ArithOps ar= new ArithOps();
+		//ArithOps ar= new ArithOps();
 		System.out.println("The given 3 numbers are: " + a +", " +b+ ", " +c);
-		double avAnswer= (ar.sum(a, b, c))/3;
+		double avAnswer= (sum(a, b, c))/3;
 		if(avAnswer >100) {
 			System.out.println("Average of given 3 values exceeds 100%, ie: " +avAnswer+ " please enter new values.");
 			return 00.0;
@@ -27,7 +27,6 @@ class ArithOps{
 		boolean boolExp = false;
 			if (avAnswer> 50){
 				boolExp= true;
-				
 			}
 			else if (avAnswer<50){
 				boolExp= false;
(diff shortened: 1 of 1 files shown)


src/vireshJJun20/ArithOps.java
```diff
@@ -0,0 +1,81 @@
...
+
+	double sum(double a, double b, double c){
+		double smAnswer;
+		smAnswer= a+b+c;
+		System.out.println(" - Sum: " + smAnswer);
+		return smAnswer;
+	}
+	
+	double average(double a, double b, double c){
+		ArithOps ar= new ArithOps();
```
KrishnaKTechnocredits: No need to create new object. non static to non static method can be called directly.

src/vireshJJun20/ArithOps.java
```diff
@@ -0,0 +1,81 @@
...
+		System.out.println(" - Average: " +avAnswer);
+		return avAnswer;
+		}
+	}
+	
+	static boolean isAnswerAboveExpecation(double avAnswer){
+		boolean boolExp = false;
+			if (avAnswer> 50){
+				boolExp= true;
+
```
KrishnaKTechnocredits: remove extra space

src/vireshJJun20/ArithOps.java
```diff
@@ -0,0 +1,81 @@
...
+	public static void main(String[] args){
+		ArithOps arithOps= new ArithOps();
+		double avg= arithOps.average(99.0,99.0,99.0);
+		if (avg!=00.0) {
+			if(ArithOps.isEligible(avg)== true) {
+				System.out.println("Is Eligible: Yes (Average is more than 50 or divisible by 2)");
+			}
+			else
+				System.out.println("Is Eligible: No (Average is less than or equals 50 or not divisible by 2)");
+				//System.out.println("Is Eligible: " + ArithOps.isEligible(avg));
```
KrishnaKTechnocredits: It should be and.
if not divisible by 2 and less than or equals 50...
System.out.println("Is Eligible: No (Average is less than or equals 50 or not divisible by 2)");

src/vireshJJun20/ArithOps.java
```diff
@@ -0,0 +1,81 @@
...
+		else{
+			grade= "Grade B";
+		}
+		return grade;
+	}	
+	
+	public static void main(String[] args){
+		ArithOps arithOps= new ArithOps();
+		double avg= arithOps.average(99.0,99.0,99.0);
+		if (avg!=00.0) {
```
KrishnaKTechnocredits: If avg is == 00.0 then we dont want to even call other methods like to getYourGrade or isEligible ?

VireshJJun20: Done with the changes

src/vireshJJun20/ArithOps.java
VireshJJun20: Done with the changes.

src/vireshJJun20/ArithOps.java
VireshJJun20: Done.

src/vireshJJun20/ArithOps.java
VireshJJun20: Removed the step 13 (object creation with ref variable ar)

PR #138, opened 2020-06-17
Madhura_Assignment_9_Done

MadhuraMJun20:
-->Static, non static concept
-->return type concept

### DIFF
--- a/src/madhuraMJun20/TestStaticNonStatic.java
+++ b/src/madhuraMJun20/TestStaticNonStatic.java
@@ -1,0 +1,51 @@
+package madhuraMJun20;
+
+public class TestStaticNonStatic {
+
+	double sum(double a, double b, double c) {
+		double sumAns= a+b+c;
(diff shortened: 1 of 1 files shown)


src/madhuraMJun20/TestStaticNonStatic.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class TestStaticNonStatic {
+
+	double sum(double a, double b, double c) {
+		double sumAns= a+b+c;
+		System.out.println("Sum of the marks is : " +sumAns);
+		return sumAns;
+	}
+	double average(double a, double b, double c) {
+		double avgAns= (a+b+c)/3;
```
KrishnaKTechnocredits: you should call sum method here.
double avgAns = sum(a,b,c)/3

MadhuraMJun20: Done.

KrishnaKTechnocredits (commented):
Now, code looks good.

PR #140, opened 2020-06-17
Pranita_Assignment_9_Done

PranitaPJun20:
Design to test static, non static, return type concept.Added methods as per requirement.

### DIFF
--- a/src/pranitaPJun20/MethodCalling.java
+++ b/src/pranitaPJun20/MethodCalling.java
@@ -1,0 +1,81 @@
+/*Assignment 9 : 17th Jun 2020
+
+Design to test static, non static, return type concept.
+
+1) Method sum(), accept 3 parameters of type double & return answer
+2) Method average(), accept 3 parameters of type double & return answer
+3) Method isAnswerAboveExpecation(), take one parameter (answer which you received from average method) and return true if value is more then 50 else false.
+4) Method isEligible(),take one parameter (answer which you received from average method) return true if value is more then 50 or divisible by 2. Eligibility will be printed in main method. 
+5) Method getYourGrade(),take one parameter (answer which you received from average method) return "A Grade" if average is more then 80 else "B grade", main method will print grade received from getYourGrade() method.
+
+Rules : a) From main method, you will call average() and average() method will call sum() method.
+               Main method will not directly call sum method. 
+	    b) average() & sum() should be non static method.
+	    c) isAnswerAboveExpecation(), isEligible(), getYourGrade() method should be static method and      called from main method.
+            d) Class should not have any instance variable.  
+			
+*/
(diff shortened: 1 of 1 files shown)


src/pranitaPJun20/MethodCalling.java
```diff
@@ -0,0 +1,81 @@
...
+	double sum(double num1,double num2,double num3){
+		double sumLocal = num1+num2+num3;
+		//System.out.println("The sum is:"+sumLocal);
+		return sumLocal;	
+	}
+//Method average(), accept 3 parameters of type double & return answer
+	double average(double num1,double num2,double num3){
+		sum(80.5,90.9,78.8);
+		double averageLocal = (num1+num2+num3)/3;
+		//System.out.println("The average is:"+averageLocal);
```
KrishnaKTechnocredits: line number 28, you should capture it in a variable and line number 30 should be.
double totalSum = sum(80.5,90.9,78.8);
double averageLocal = totalSum/3

src/pranitaPJun20/MethodCalling.java
```diff
@@ -0,0 +1,81 @@
...
+		return averageLocal;
+	}
+//Method isAnswerAboveExpecation(), take one parameter (answer which you received from average method)
+// and return true if value is more then 50 else false.
+	static boolean isAnswerAboveExpectation(){
```
KrishnaKTechnocredits: You should pass argument here. You have already calculated average once, you should use that value in rest method by passing it as parameter.

src/pranitaPJun20/MethodCalling.java
```diff
@@ -0,0 +1,81 @@
...
+		double avg= methodCalling3.average(100,110,120);
+			if(avg>80)
+				return"A Grade";
+			else
+				return "B Grade";			
+		}
+		
+	public static void main(String[] args){
+		MethodCalling methodCalling = new MethodCalling();
+		methodCalling.average(100,90.99,78.89);
```
KrishnaKTechnocredits: You should capture return value here and use it in isAnswerAboveExpectation() and isEligible() and getYourGrade() method

src/pranitaPJun20/MethodCalling.java
PranitaPJun20: Done

src/pranitaPJun20/MethodCalling.java
PranitaPJun20: Done

src/pranitaPJun20/MethodCalling.java
PranitaPJun20: Done

KrishnaKTechnocredits (commented):
Code looks good to me.

PR #141, opened 2020-06-17
Swapna_Assignment_9_Done

SwapnaWJun20:
Swapna_Assignment_9_Done for 17/06/2020

### DIFF
--- a/src/swapnaWJun20/Calculations.java
+++ b/src/swapnaWJun20/Calculations.java
@@ -1,0 +1,78 @@
+package swapnaWJun20;
+/*Assignment 9 : 17th Jun 2020
+
+Design to test static, non static, return type concept.
+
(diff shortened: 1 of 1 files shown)


src/swapnaWJun20/Calculations.java
```diff
@@ -0,0 +1,78 @@
...
+
+	static boolean isEligible(double avg){
+		if(avg>50|| avg%2==0)
+			return true;
+		else
+			return false;
+	}
+
+	static String getYourGrade(double avg){
+		if (avg>=80)
```
KrishnaKTechnocredits, line 52: 5) Method getYourGrade(),take one parameter (answer which you received from average method)
 return "A Grade" if average is more then 80 else "B grade", main method will
  print grade received from getYourGrade() method.

It should be if(avg>80) only

PR #142, opened 2020-06-17
Barkha_Assignment_9_Done

BarkhaPJun20:
Design to test static, non static, return type concept.

1) Method sum(), accept 3 parameters of type double & return answer
2) Method average(), accept 3 parameters of type double & return answer
3) Method isAnswerAboveExpecation(), take one parameter (answer which you received from average method) and return true if value is more then 50 else false.
4) Method isEligible(),take one parameter (answer which you received from average method) return true if value is more then 50 or divisible by 2. Eligibility will be printed in main method. 
5) Method getYourGrade(),take one parameter (answer which you received from average method) return "A Grade" if average is more then 80 else "B grade", main method will print grade received from getYourGrade() method.

Rules : a) From main method, you will call average() and average() method will call sum() method.
               Main method will not directly call sum method. 
	    b) average() & sum() should be non static method.
	    c) isAnswerAboveExpecation(), isEligible(), getYourGrade() method should be static method and      called from main method.
            d) Class should not have any instance variable.

### DIFF
--- a/src/barkhaPJun20/StaticNonstatic.java
+++ b/src/barkhaPJun20/StaticNonstatic.java
@@ -1,14 +1,12 @@
 package barkhaPJun20;
 
 public class StaticNonstatic {
-	 double sum( double x, double y, double z){
-		return (x+y+z);
+	 double sum( double num1, double num2, double num3){
+		return (num1+num2+num3);
 	}	 
-	  double average(double x1, double y1,double z1){
-		StaticNonstatic s2=new StaticNonstatic();
-		double totalAvg= (s2.sum(x1,y1,z1)/3);
-		return totalAvg;
-	}
+	  double average(double no1, double no2,double no3){
+		 return (sum(no1,no2,no3)/3);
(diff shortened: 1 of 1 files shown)


src/barkhaPJun20/StaticNonstatic.java
```diff
@@ -0,0 +1,46 @@
+package barkhaPJun20;
+
+public class StaticNonstatic {
+	 double sum( double x, double y, double z){
+		return (x+y+z);
+	}	 
+	  double average(double x1, double y1,double z1){
+		StaticNonstatic s2=new StaticNonstatic();
```
KrishnaKTechnocredits: Non static to non static method.. object is not required. please call directly.

BarkhaPJun20: Suggested changes are done

KrishnaKTechnocredits (commented):
Now code looks good to me.

PR #143, opened 2020-06-17
Aavruti Assignment 9

AavrutiDJun20:
Calculating Average of 3 marks and verifying average marks is above expectation or not, is eligible or not and Grade of the student.


KrishnaKTechnocredits (approved):
Code looks good to me. It is as per given requirement. Keep it up.

PR #144, opened 2020-06-17
Raj_Assignment_9_Done

RajAJun20:
Static and NonStatic method implementation

### DIFF
--- a/src/rajAJun20/StaticNonStatic.java
+++ b/src/rajAJun20/StaticNonStatic.java
@@ -1,0 +1,41 @@
+package rajAJun20;
+
+public class StaticNonStatic {
+			double sum(double var1, double var2, double var3) {
+	    	double sum = var1 + var2 + var3;
+	    	System.out.println("Sum of three variable is: " + sum);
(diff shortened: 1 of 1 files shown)


src/rajAJun20/StaticNonStatic.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class StaticNonStatic {
+			double sum(double var1, double var2, double var3) {
+	    	double sum = var1 + var2 + var3;
+	    	System.out.println("Sum of three variable is: " + sum);
+	    	return sum;
+	    }
+	    double average(double var1, double var2, double var3) {
+	    	double average = (var1 + var2 + var3)/3;
+	    	sum(var1, var2, var3);
```
KrishnaKTechnocredits: no need of line 11.
line number 10 is not correct. you should write
double average = sum(var1 ,var2,var3)/3;

RajAJun20: Done Suggested changes....!!!

PR #148, opened 2020-06-17
Megha_Assignment_9Done

MeghaJJune20:
-->StaticNonStaticReturnType Assignment completed

### DIFF
--- a/src/meghaJJune20/Number.java
+++ b/src/meghaJJune20/Number.java
@@ -11,7 +11,7 @@ public class Number {
 	}
 	double average(double x,double y,double z)
 	{
-		double avg=(x+y+z)/3;
+		double avg= sum(x,y,z)/3;
 		System.out.println("Total Average is : " +avg);
 		return avg;
 	}
@@ -35,7 +35,7 @@ public class Number {
 	static String getYourGrade(double avg)
 	{
 		System.out.println("Student Grade ");
-		if (avg >50)
+		if (avg >80)
 			return "Grade A";
 		else
 			return "Grade B";
@@ -43,7 +43,7 @@ public class Number {
 	public static void main(String [] args)
 	{
 		Number number = new Number();
-		double avg=number.average(40,50,50);
+		double avg=number.average(80,50,30);
 		
 		System.out.println();
 		System.out.println(isAnswerAboveExpecation(avg));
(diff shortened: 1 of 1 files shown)


src/meghaJJune20/Number.java
```diff
@@ -0,0 +1,57 @@
...
+	double sum(double x,double y,double z)
+	{
+		double totalsum=x+y+z;
+		System.out.println("Sum Total is : " +totalsum);
+		return totalsum;
+		
+	}
+	double average(double x,double y,double z)
+	{
+		double avg=(x+y+z)/3;
```
KrishnaKTechnocredits: you should say averageAnswer = sum(x,y,z)/3; Otherwise there is no need of sum method. we asked you call sum method from average. Please correct it.

src/meghaJJune20/Number.java
```diff
@@ -0,0 +1,57 @@
...
+			System.out.println("Eligible ");
+			if(avg>50 || avg%2==0)
+				return true;
+			else
+				return false;
+	}
+	static String getYourGrade(double avg)
+	{
+		System.out.println("Student Grade ");
+		if (avg >50)
```
KrishnaKTechnocredits: It should be more then 80. Please check requirement.

src/meghaJJune20/Number.java
MeghaJJune20: Updated as per comments

src/meghaJJune20/Number.java
MeghaJJune20: Updated as per comments

src/meghaJJune20/Number.java
```diff
@@ -0,0 +1,57 @@
...
+	}
+	public static void main(String [] args)
+	{
+		Number number = new Number();
+		double avg=number.average(80,50,30);
+		
+		System.out.println();
+		System.out.println(isAnswerAboveExpecation(avg));
+		System.out.println();
+		System.out.println(isEligible(avg));
```
KrishnaKTechnocredits: Still not as per expectation, Review #170 PR.

MeghaJJune20: --> Changes done as per comments.

KrishnaKTechnocredits (commented):
Code looks good.

PR #149, opened 2020-06-17
Palak Assignment 9

PalakSJun20:


src/palakSJun20/StaticNonStaticConcept.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class StaticNonStaticConcept {
+	
+	double sum(double number1, double number2, double number3){
+		return number1 + number2 + number3;
+	}
+	
+	double average(double number1, double number2, double number3){
+		double averageResult=(number1 + number2 + number3) / 3;
+		sum(10.0, 10.5, 10.5);
```
KrishnaKTechnocredits: you should say average = sum(variable1,variable2,variable3)/3; Otherwise there is no need of sum method. we asked you to call sum method from average. Please correct it.

PalakSJun20: Updated Code as per your comments.

KrishnaKTechnocredits (commented):
Code looks good to me.

PR #157, opened 2020-06-17
Yash_Assignment_9_Done

YashCJun20:
Assignment Completed.
Please review

### DIFF
--- a/src/yashCJun20/MathOperation.java
+++ b/src/yashCJun20/MathOperation.java
@@ -55,7 +55,7 @@ class MathOperation{
 		
 	 MathOperation mathOperation = new MathOperation();
 	 double ans=mathOperation.average(20.00,30.00,40.00);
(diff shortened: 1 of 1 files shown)


src/yashCJun20/MathOperation.java
```diff
@@ -0,0 +1,76 @@
...
+			return "A Grade";
+		else
+			return "B Grade";
+	}
+	
+	public static void main(String[] args){
+		
+	 MathOperation mathOperation = new MathOperation();
+	 double ans=mathOperation.average(20.00,30.00,40.00);
+	 isAnswerAboveExpecation(ans);
```
KrishnaKTechnocredits: You should print it, otherwise How we would know ?

YashCJun20: Done

KrishnaKTechnocredits (approved):
It's better now.

PR #158, opened 2020-06-17
Harshad_Assignment_09_Done

HarshadBJun20:
->Assignment 09 Done with use of static, non-static and return type.

### DIFF
--- a/src/harshadBJun20/TestStaticNonStaticReturnType.java
+++ b/src/harshadBJun20/TestStaticNonStaticReturnType.java
@@ -1,0 +1,67 @@
+/*Assignment 9 : 17th Jun 2020
+
+Design to test static, non static, return type concept.
+
+1) Method sum(), accept 3 parameters of type double & return answer
+2) Method average(), accept 3 parameters of type double & return answer
+3) Method isAnswerAboveExpecation(), take one parameter (answer which you received from average method) and return true if value is more then 50 else false.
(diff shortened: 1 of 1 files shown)


src/harshadBJun20/TestStaticNonStaticReturnType.java
```diff
@@ -0,0 +1,67 @@
...
+		double avgAnswer = test.average(100,80,50);
+		//System.out.println("Value of isAnswerAboveExpecation: "+isAnswerAboveExpecation(avgAnswer));
+		
+		if(isEligible(avgAnswer)== true)
+			System.out.println("Student with "+avgAnswer+" average is Eligibe");
+		else
+			System.out.println("Student with "+avgAnswer+" average is not Eligibe");
+		
+		System.out.println("Grade received: "+getYourGrade(avgAnswer));	
+	}
```
KrishnaKTechnocredits, line 66: isAnswerAboveExpecation() You forgot to call this method. Else code looks perfect.

HarshadBJun20: ->Actually, I have already called the isAnswerAboveExpecation(ans) method in if condition of isEligible(ans).
->Now I have printed the return value also in main method.
->Done.

KrishnaKTechnocredits (commented):
Code looks good to me.

PR #174, opened 2020-06-18
Aavruti Assignment 10 Combination of Multiple Classes

AavrutiDJun20:
Combined combination of static and non-static methods of multiple classes.


src/aavrutiDJun20/FourthClass.java
```diff
@@ -0,0 +1,14 @@
+package aavrutiDJun20;
+
+public class FourthClass {
+
+	static void firstClassM1() {
```
KrishnaKTechnocredits, line 5: Here it should be fourthClassM1(), the remaining code looks good to me.

PR #196, opened 2020-06-18
Sweety_Assignment_10

SweetyMJun20:
Please review Assignment 10


src/sweetyMJun20/ClientClass.java
```diff
@@ -0,0 +1,24 @@
+package sweetyMJun20;
+
+public class ClientClass {
+	
+	public static void main(String[] args){
+		FirstClass firstclass=new FirstClass();
```
KrishnaKTechnocredits, line 7: Code is not as per requirement.

PR #205, opened 2020-06-18
ADT_Assignment_10 Done

AditiGJUN20:
// Code to check communication between class and how to cal static and non-static methods between the classes


KrishnaKTechnocredits (commented):
Good job, Code looks good to me.

PR #212, opened 2020-06-19
Aastha_Assignment_11_Done

AasthaJJun20:
-> Created 2 classes : 1. with all the operation
                                   2. for Client
-> Used Switch in method bankOperation() for credits/Debits in particular bank
-> Used Switch in methods displayIndividualBankCredits & displayIndividualBankDebits to display the total credit/Debit operations performed on a particular bank
-> default is also given in methods displayIndividualBankCredits & displayIndividualBankDebits if user wants the total credit/Debit operations performed on all the banks
->PFA ouput's screenshot

MoneyManagement-output

### DIFF
--- a/src/aasthaJJun20/MoneyManagement.java
+++ b/src/aasthaJJun20/MoneyManagement.java
@@ -1,0 +1,173 @@
+package aasthaJJun20;
+
+public class MoneyManagement {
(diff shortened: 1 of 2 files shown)


KrishnaKTechnocredits (commented):
Good job Aastha, I like the way you have written.

PR #223, opened 2020-06-19
Pranita_Assignment_11_Done

PranitaPJun20:
Total balance value is printed using values returned from switch case statement.
Created separate method to print bank response message.
SwitchCase


KrishnaKTechnocredits (approved):
@PranitaPJun20 Code looks good to me. Keep it up.

PR #228, opened 2020-06-19
Assignment Done

PiyushPJun20:


src/BankingSolution.java
```diff
@@ -0,0 +1,147 @@
+
+public class BankingSolution {
+	static double iciciBal, hdfcBal, bobBal, sbiBal;
+	static int totalCreditCnt, totalDebitCnt;
+	static int iciciCrCount, iciciDbCount, hdfcCrCount, hdfcDbCount, bobCrCount, bobDbCount, sbiCrCount, sbiDbCount;
```
KrishnaKTechnocredits, line 5: For every variable, we should not create static variable.

KrishnaKTechnocredits (commented):
Code looks, good to me. take care from next time.

PR #229, opened 2020-06-19
Assginment 11 completed

KARAN1595:
Completed Assginment 11

### DIFF
--- a/src/kARAN1595/BankAccount.java
+++ b/src/kARAN1595/BankAccount.java
@@ -1,8 +1,4 @@
-package kARAN1595;
-
-public class BankAccount {
-
-	/*Assignment 11 :  19th Jun 2020
+/*Assignment 11 :  19th Jun 2020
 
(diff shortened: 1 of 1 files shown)


src/kARAN1595/BankAccount.java
```diff
@@ -0,0 +1,264 @@
...
+
+public class BankAccount {
+
+	
+	
+	int iciciCreditCount,iciciDebitCount;
+	int hdfcCreditCount,hdfcDebitCount;
+	int sbiCreditCount,sbiDebitCount;
+	int bobCreditCount,bobDebitCount;
+	int totalCreditCount,totalDebitCount;
```
KrishnaKTechnocredits, line 39: For this two-variable you should create static variable. Because right now you are performing an action using one reference variable (means for one user) what about if you want to perform for 10 users.

KrishnaKTechnocredits (commented):
Except that commented code looks good to me.

PR #249, opened 2020-06-20
Aavruti Assignment_12 Scanner

AavrutiDJun20:
Performing Bank Operations like Debit, Credit, print Bank Balance using Scanner.
Scanner2
Scanner1


KrishnaKTechnocredits (commented):
Good job dear, Keep it up.

PR #263, opened 2020-06-20
Pranita_Assignment_12_Done

PranitaPJun20:
Used scanner to take input from user
Assign_11_scanner1
Assign_12_Scanner

### DIFF
--- a/src/pranitaPJun20/BankingOperationScanner.java
+++ b/src/pranitaPJun20/BankingOperationScanner.java
@@ -1,0 +1,219 @@
+package pranitaPJun20;
+
+import java.util.Scanner;
+
+public class BankingOperationScanner {
+	static int staticDebitCount,staticCreditCount;
+	double balanceIcici,balanceSbi,balanceHdfc,balanceBob;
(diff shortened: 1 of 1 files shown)


src/pranitaPJun20/BankingOperationScanner.java
```diff
@@ -0,0 +1,219 @@
...
+		}
+		else {
+			return "*Please spend money wisely*";
+		}
+	}
+	
+	public static void main(String[] args) {
+		BankingOperationScanner bankingOperationScanner = new BankingOperationScanner();
+		Scanner scanner = new Scanner(System.in);
+		for(int i=1;i<=2;i++) {
```
KrishnaKTechnocredits: Any specific reason of writing for a loop.

PranitaPJun20: >Actually 4 credit debit operations were there so i thought to use for loop.But i will update the code with do while loop.Will write it properly.

KrishnaKTechnocredits (approved):
@PranitaPJun20 Code looks good to me. Keep it up.

PR #291, opened 2020-06-21
Aastha_Assignment_13_Done_54Mins

AasthaJJun20:
-> pushed 2 class files: FindMaxMinNumber and LongestName
-> Arrays are used and inputs are taken from user
-> input for - size of array, is also taken from user
-> Handled the case if the user accidentally enters a negative value or 0 as the size of array 
-> PFB output
FindMaxMinNumber-output
LongestName-output

### DIFF
--- a/src/aasthaJJun20/FindMaxMinNumber.java
+++ b/src/aasthaJJun20/FindMaxMinNumber.java
@@ -1,0 +1,55 @@
+package aasthaJJun20;
+import java.util.Scanner;
+
+public class FindMaxMinNumber {
+	
+	int maxNumber(int[] numbers) {
+		int max = numbers[0];
+		for(int index = 1; index < numbers.length; index++) {
+			if(numbers[index] > max)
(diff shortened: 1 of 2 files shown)


src/aasthaJJun20/FindMaxMinNumber.java
```diff
@@ -0,0 +1,55 @@
...
+				min = numbers[index];
+		}
+		return min;
+	}
+	
+	public static void main(String[] args) {
+		FindMaxMinNumber findMaxMinNumber = new FindMaxMinNumber();
+		Scanner scanner = new Scanner(System.in);
+		String status = "";
+
```
KrishnaKTechnocredits, line 28: remove extra blank line.

KrishnaKTechnocredits (approved):
Code looks good to me Aastha. You are doing good. Lets try to reduce time. It should not be more than 30 mins.

PR #301, opened 2020-06-21
Barkha_Assignment_13

BarkhaPJun20:
Program 1 :
A) Find maximum number from user given array [use scanner class to take numbers and size of an array]
B) Find minimum number from user given array [use scanner class to take numbers and size of an array]

Assigment 2 :
A) Find the name having maximum length from user given array [use scanner class to take names and size of an array].
image
image
Issue: Not getting why longest Name is coming null.
Also, its confusing how to use scanner class. Need more practice on Array.

### DIFF
--- a/src/barkhaPJun20/MaxLength.java
+++ b/src/barkhaPJun20/MaxLength.java
@@ -1,0 +1,43 @@
+package barkhaPJun20;
+
+import java.util.Scanner;
+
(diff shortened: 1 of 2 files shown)


KrishnaKTechnocredits (approved):
Good job dear, From next time onwards please add proper comments on methods.

PR #304, opened 2020-06-21
Palak Assignment 13_50 Min

PalakSJun20:
1. In Name Length Program what if we enter same length name then how can we print all Name as of now last name is printing if length is name. 

**Note :** Committed Morning Program as well. 
image
image


KrishnaKTechnocredits (approved):
Good job. From next time add comments on before every method.

PR #315, opened 2020-06-21
Assignment Done

PiyushPJun20:
Assignment_Done_1Hour

### DIFF
--- a/src/MaxMin.java
+++ b/src/MaxMin.java
@@ -1,0 +1,41 @@
+import java.util.Scanner;
+
+public class MaxMin {
+	
+	void maximum(int[] input) {
+		int max = input[0];
+		for(int index=0; index<input.length; index++) {
+			if(input[index]>max)
+				max = input[index];
+		}
+		System.out.println(" Maximum Number is: " + max);
+	}
+	
+	void minimum(int[] input) {
+		int min = input[0];
+		for(int index=0; index<input.length; index++) {
+			if(input[index]<min)
+				min = input[index];
+		}
+		System.out.println(" Minimum Number is: " + min);
+	}
+
+	public static void main(String[] args) {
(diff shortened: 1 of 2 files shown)


src/MaxMin.java
```diff
@@ -0,0 +1,41 @@
+import java.util.Scanner;
+
+public class MaxMin {
+	
+	void maximum(int[] input) {
+		int max = input[0];
+		for(int index=0; index<input.length; index++) {
```
KrishnaKTechnocredits: input[0] is already in max. you should start loop from index 1

src/MaxMin.java
```diff
@@ -0,0 +1,41 @@
...
+			if(input[index]>max)
+				max = input[index];
+		}
+		System.out.println(" Maximum Number is: " + max);
+	}
+	
+	void minimum(int[] input) {
+		int min = input[0];
+		for(int index=0; index<input.length; index++) {
```
KrishnaKTechnocredits: same here.

src/NameLength.java
```diff
@@ -0,0 +1,32 @@
+import java.util.Scanner;
+
+public class NameLength {
+	
+	void printLength(String[] input) {
+		String max = input[0];
+		for(int index=0; index<input.length;index++) {
```
KrishnaKTechnocredits: same here.

src/MaxMin.java
PiyushPJun20: Changes Incorporated.

src/MaxMin.java
PiyushPJun20: Changes Incorporated.

src/NameLength.java
PiyushPJun20: Changes Incorporated. Thanks.

PR #320, opened 2020-06-22
Ajit_Assignment_14_35mins

AjitSJune20:
image


KrishnaKTechnocredits (approved):
Code looks good to me. Good job. Like the both for loop int variable name.

PR #323, opened 2020-06-22
Aastha_Assignment_14_Done_45mins

AasthaJJun20:
-> pushed FrequencyOfChar.java file for assignment 14
->Handled the case to ignore the case of entered names[ ] as well as the entered char
->It took me approx 45 mins to complete, since I was struggling to use .equalIgnoreCase( ) method but its done now.
->PFB screenshot:
FrequencyOfChar-ofEachStringInAnArray-output


KrishnaKTechnocredits (approved):
Good Job Aastha, Like the way you represent the console o/p.

PR #340, opened 2020-06-22
Raj_Assignment_14_Done

RajAJun20:
Please enter size of an array: 3
Enter character to find its frequency : a
Enter name 1:Raj
Enter name 2:Adatiya
Enter name 3:Akshay
Frequency of a in strings of array is : 4


KrishnaKTechnocredits (approved):
Code looks good to me.

PR #342, opened 2020-06-22
Archana Assignment 14_20mins

ArchanaNJun20:
image


ArchanaNJun20:
Hi Krishna,
I feel like i am taking extra time to build correct logic.however after daily practice i will able to reduce that time.hope i am right?
Thank you

KrishnaKTechnocredits (approved):
Code looks good to me. And dear 20 mins is not extra time. it is the expected time for this movement. Don't worries you are on track. keep practicing.

PR #356, opened 2020-06-23
Barkha_Assignment_15_35mins

BarkhaPJun20:
Program 1 : 
    You need to store user input in array and do sum of even numbers only from created 
     array. [ideal Expected Time : 10 mins, Max Time : 15 mins]

     Example :
     Input : {10,5,7,3,1,5,9,30}
     Output: 40

Program 2 : 
    Create a class where user will provide int array. *[ideal Expected Time : 15 mins, Max 
    Time : 25 mins]*

    A) User can provide any positive and negative number

    B) You need to find out how many positive and negative numbers are there in array 
    (Hint : here you need two variable which maintains count of positive and negative 
    number)
    Example : input -> 10,-2,-3,4,7
    Output -> 3 positive & 2 negative numbers

    C) Create a method which return difference of positive number sum and negative number 
     sum.
     Example : input -> 10,-2,-3,4,7
     Output -> 16
Program 1:
image
Program 2:
image

### DIFF
--- a/src/barkhaPJun20/AdditionOfEven.java
+++ b/src/barkhaPJun20/AdditionOfEven.java
@@ -1,0 +1,37 @@
+package barkhaPJun20;
+
+import java.util.Scanner;
+
+public class AdditionOfEven {
+	
+	int sum=0;
+	void sumOfEvenNum(int[] num) {
+		for (int i=0; i<num.length; i++) {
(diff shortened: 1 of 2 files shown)


KrishnaKTechnocredits (approved):
Good job Barkha, Keep it up.

PR #364, opened 2020-06-23
Shruti_Assignment 15_1 hour

ShrutiDJun20:
Output of program-Print sum of even numbers:
image

Output of program- Difference between positive and negative:
image


KrishnaKTechnocredits (approved):
Code looks good to me Keep it up, dear. need to work on time.

PR #365, opened 2020-06-23
Aavruti_Assignment_15_16mins

AavrutiDJun20:
Program 1: Sum of Even Number provided by user in array
Program 2: i) Count of Negative and Positive Number
ii) Sum of Positive and Negative Number

Prog1_15

Prog2_15


KrishnaKTechnocredits (approved):
Code looks good to me Keep it up, dear.

PR #379, opened 2020-06-23
ADT_Assignment_15_Done_30Mins

AditiGJUN20:
- Code is pushed for Assignment_15
1. **SumOfEvenNum.java** - Takes user input for array and does sum of even numbers.
Screen Shot 2020-06-23 at 10 32 25 AM

2. **PositiveNegativeIntOps.java** - 
- Takes user input for array 
- Count positive and negative integers (Added condition to check unsigned integer zero and notifying user if there is any in input) 
- Method returns difference of sum of positive and negative integers
Screen Shot 2020-06-23 at 10 33 56 AM


KrishnaKTechnocredits (approved):
Good job dear, keep it up.

PR #392, opened 2020-06-24
Barkha_Assignment_16_>2 hrs

BarkhaPJun20:
1) Create a class where you need to check given number is there in predefined array or not, if present then print frequency of the number. [Estimated time: 15 mins]

Input : Predefined array -> {10,3,9,-99,81,3,-11}
           Number to be check -> 33
Output : Given number 33 is not present in predefined array.

Input : Number to be check -> 3
output : Given number 3 is present in predefined array with frequency 2.

2) Create a class with following functionality. [Estimated time: 35 mins]
a) Find one character frequency from given name.
     Input: name - technocredits
               character - e
     Output : e -> 2 time in technocredits

b) Find all character frequency from given name.
     Input: name - aakanksha
     Output : a -> 4
	           a -> 4
                   k -> 2
                   a -> 4
                   n -> 1
                   k -> 2
                   s -> 1
                   h -> 1
                   a -> 4
c) Find all character frequency from each name given in String array.
     Input : Array -> {"raj", "aarya", "aavruti", "shruti"}
     Output : r -> 1 time in raj
                   a -> 1 time in raj
                   j -> 1 time in raj
         =====================
		   a -> 3 time in aarya
		   a -> 3 time in aarya
		   r -> 1 time in aarya
                   y -> 1 time in aarya
                   a -> 3 time in aarya	
         ====================
          And so on....		 
Program 1: Not getting exact output(tried with if else as well) plz suggest
image

Program 2: 
image

### DIFF
--- a/src/barkhaPJun20/AllCharInArray.java
+++ b/src/barkhaPJun20/AllCharInArray.java
@@ -1,0 +1,55 @@
+package barkhaPJun20;
+
+import java.util.Scanner;
+
+public class AllCharInArray {
+	
+	void findCharFreq(String name, char ch) {
+		int count=0;
+		for (int index=0;index<name.length();index++) {
+			if (name.charAt(index)==ch)
+				count++;
+		}	
+		System.out.println("Number of character "+ch+" in "+"name "+" is "+count);
+		
+	}	
+	void findAllCharFreq(String name1) {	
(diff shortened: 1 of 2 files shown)


KrishnaKTechnocredits (approved):
The code looks good to me. Good job.

PR #399, opened 2020-06-24
Palak Assignment 16_70Min

PalakSJun20:
image
image
image


src/palakSJun20/ArrayFreqOfNumber.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class ArrayFreqOfNumber {
+
+	void frequencyOfNumber(int[] array , int number){
+		int count = 0 ;
+		for (int index = 0; index < array.length; index++) {
+			if(array[index] == number) 
+				count++;	
+		}
+			if(count == number)
```
KrishnaKTechnocredits: Enter the number you want to search in array
33
Given Number 33 is present in predefined array with frequency 0

Your output is wrong, Please check this condition. if(count == number) instead of if(count==0) will works. 
humane code a6e se test nahi kiya hai

PalakSJun20: i updated the condition while attaching the SS but missed to update here. 
Condition updated.

KrishnaKTechnocredits: okay.

PR #422, opened 2020-06-25
Nikhil_Assignment_17_Done_(10+20)30Min

NikhilMJun20:
->Completed and pushed Assignment # 17.
->Please review and comment.

**Observation** : 
Realisation of condition for 2nd program took almost 15 min. Need to think quickly about such conditions. O/p is supposed to be 't e c'. My code was working fine but, o/p was 'c e t'.

->Screenshots of the outputs:-

Output_CharClassify

Output_FirstNLastIndexOfChar

### DIFF
--- a/src/nikhilMJun20/CharClassify.java
+++ b/src/nikhilMJun20/CharClassify.java
@@ -1,0 +1,39 @@
+/*Count Digits,Letters, UpperCase characters, LowerCase characters, Special characters from predefined string.
+
+Input: 1rRpd3F9#K(E
+Output : 
+Total digit: 3
+Total Letters: 7
+Total Uppercase: 4
+Total Lowercase: 3
+Total Special char: 2
+*/
+
+package nikhilMJun20;
+
+public class CharClassify {
+	void findCharType(String str) {
+		int digitCnt=0, letterCnt=0, upperCnt=0, lowerCnt=0, specialCnt=0;
+		
+		for(int index=0; index<str.length(); index++) {
+			if(Character.isDigit(str.charAt(index))) {
+				digitCnt++;
+			}else if(Character.isLetter(str.charAt(index))) {
+				letterCnt++;
+				if(Character.isLowerCase(str.charAt(index)))
+					lowerCnt++;
+				else
+					upperCnt++;
+			}else
+				specialCnt++;
+		}
+		System.out.println("String Given: "+str+"\n\nTotal Digits: "+digitCnt+"\nTotal Letters: "+letterCnt+"\nTotal Lower Case Letters: "+lowerCnt+"\nTotal Upper Case Letters: "+upperCnt+"\nTotal Special Characters: "+specialCnt);
+	}
+	
+	public static void main(String[] args) {
+		CharClassify charClassify = new CharClassify();
+		
+		String str = "1rRpd3F9#K(E";
+		charClassify.findCharType(str);
+	}
+}
--- a/src/nikhilMJun20/CharClassify.java
+++ b/src/nikhilMJun20/CharClassify.java
@@ -13,21 +13,21 @@ package nikhilMJun20;
 
 public class CharClassify {
 	void findCharType(String str) {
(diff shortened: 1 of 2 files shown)


src/nikhilMJun20/CharClassify.java
```diff
@@ -0,0 +1,39 @@
...
+package nikhilMJun20;
+
+public class CharClassify {
+	void findCharType(String str) {
+		int digitCnt=0, letterCnt=0, upperCnt=0, lowerCnt=0, specialCnt=0;
+		
+		for(int index=0; index<str.length(); index++) {
+			if(Character.isDigit(str.charAt(index))) {
+				digitCnt++;
+			}else if(Character.isLetter(str.charAt(index))) {
```
KrishnaKTechnocredits: From line number 21 to 29, you could write some better logic, think about it. Or will let you know. No need to check else if and if conditions and letterCnt this variable is also not needed.

src/nikhilMJun20/FirstNLastIndexOfChar.java
```diff
@@ -0,0 +1,33 @@
...
+
+package nikhilMJun20;
+
+public class FirstNLastIndexOfChar {
+	void displayFirstLastIndex(String str) {
+		System.out.println("Input : "+str+"\nOutput:-");
+		
+		//This detects fist and last occurrence but, in reverse order i.e. c e t
+		for(int index=0; index<str.length(); index++) {
+			if(index == str.indexOf(str.charAt(index)) && str.indexOf(str.charAt(index)) < str.lastIndexOf(str.charAt(index))) {
```
KrishnaKTechnocredits, line 21: This method looks good to me.

src/nikhilMJun20/CharClassify.java
NikhilMJun20: Hi Krishna, 
I have modified the code.
But, please do let me know if there's better logic than this. I will modify accordingly.
Thanks!

src/nikhilMJun20/CharClassify.java
```diff
@@ -13,21 +13,21 @@
...
 					lowerCnt++;
 				else
 					upperCnt++;
-			}else
+				}
+			else
 				specialCnt++;
 		}
-		System.out.println("String Given: "+str+"\n\nTotal Digits: "+digitCnt+"\nTotal Letters: "+letterCnt+"\nTotal Lower Case Letters: "+lowerCnt+"\nTotal Upper Case Letters: "+upperCnt+"\nTotal Special Characters: "+specialCnt);
+		System.out.println("String Given: "+str+"\n\nTotal Digits: "+digitCnt+"\nTotal Letters: "+(lowerCnt+upperCnt)+"\nTotal Lower Case Letters: "+lowerCnt+"\nTotal Upper Case Letters: "+upperCnt+"\nTotal Special Characters: "+specialCnt);
```
KrishnaKTechnocredits, line 28: That logic is as expected. Good job.

src/nikhilMJun20/CharClassify.java
```diff
@@ -0,0 +1,39 @@
...
+public class CharClassify {
+	void findCharType(String str) {
+		int digitCnt=0, upperCnt=0, lowerCnt=0, specialCnt=0;
+		
+		for(int index=0; index<str.length(); index++) {
+			if(Character.isDigit(str.charAt(index)))
+				digitCnt++;
+			else if(Character.isLetter(str.charAt(index))) {
+				if(Character.isLowerCase(str.charAt(index)))
+					lowerCnt++;
```
KrishnaKTechnocredits, line 22: But here as I said you don't need to right else if and if and else block. 
if(Character.isDigit(ch))
				dCount++;
			else if(Character.isUpperCase(ch))
				uCount++;
			else if(Character.isLowerCase(ch))
				lcCount++;
			else
				sCount++;
No need to check for letters, because it will aready going to cover under uppercase and lowercase.

NikhilMJun20: Hi Krishna,
This is the most elegant solution.
Thank you for a very thorough review and comments for improvement.
I have made the changes. Please review and merge. Thanks!

PR #428, opened 2020-06-25
Apurva _assignment _16_done

ApurvaJun20:


src/apurvaJun20/Checknumber.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class Checknumber {
+	void checkgivennumber(int[] input, int num) {
+		 int count=0;
+			for(int index=0;index<input.length;index++) {
+				if(input[index]==num) {
+					count++;
+					}
+			}
+			if(count>0)
```
KrishnaKTechnocredits, line 13: No need to use count variable, you can generate output without it. try it.

src/apurvaJun20/FindCharacter.java
```diff
@@ -0,0 +1,42 @@
...
+		for(int index=0;index<name.length();index++) {
+			findonechar(name,name.charAt(index));
+		}	
+	}
+	void findeachString(String[] name)
+	{ 
+		for(int index=0;index<name.length;index++)
+		{
+		allcharfind(name[index]);
+		}
```
KrishnaKTechnocredits, line 23: i want every operation in one method only. Please make it happen.

ApurvaJun20: Could you explain this to me as how it should be

PR #431, opened 2020-06-25
Pranita_Assignment_17_Done_45min

PranitaPJun20:
image
counting

### DIFF
--- a/src/pranitaPJun20/IndexOfChar.java
+++ b/src/pranitaPJun20/IndexOfChar.java
@@ -1,0 +1,34 @@
+/* Assignment 17 : 25th Jun 2020
+
+ * 2) Write a program which print first and last index of every characters only if character has multiple occurrence.
+
+Input : technocredits
+Output :
+t -> first index = 0 Last index = 11
+e -> first index = 1 Last index = 8
+c -> first index = 2 Last index = 6
+
+Hint : To find first index of character --> indexOf() method
+            To find last index of character --> lastIndexOf() method.
+ */
+package pranitaPJun20;
+
+public class IndexOfChar {
+	
+	void displayFirstLastIndexCharString(String input) {
+		for(int index=0;index<input.length();index++) {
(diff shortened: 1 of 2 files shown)


src/pranitaPJun20/IndexOfChar.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package pranitaPJun20;
+
+public class IndexOfChar {
+	
+	void displayFirstLastIndexCharString(String input) {
+		for(int index=0;index<input.length();index++) {
+			char ch = input.charAt(index);			
+			if(index==input.indexOf(ch)) {
+				if (input.indexOf(ch)!=input.lastIndexOf(ch))
```
KrishnaKTechnocredits, line 22: Perfect logic.

src/pranitaPJun20/StringCount.java
```diff
@@ -0,0 +1,57 @@
...
+		
+		for(int index=0; index<input.length(); index++) {
+			char ch = input.charAt(index);
+			if(Character.isDigit(ch)) 
+				countDigit++;
+			
+			else if(Character.isLetter(ch)) {
+				countLetter++;
+				if(Character.isUpperCase(ch))
+					countUppercase++;
```
KrishnaKTechnocredits: else if under> if and else block not required to implement letters logic, Think here. Also countLetter without this variable you can achieve output.

src/pranitaPJun20/IndexOfChar.java
PranitaPJun20: Thank you :)

src/pranitaPJun20/StringCount.java
PranitaPJun20: Yes krishna.Got it.Will update and commit

PranitaPJun20: Changes done.Please review

PR #432, opened 2020-06-25
AMITA_Assignment_17_Done_30Mins

AmitaRJun20:
AMITA_Assignment_17_Done_30Mins....Please Review...
CharMethod
IndexOfChar

### DIFF
--- a/src/amitaRJun20/CheckCharacterMethods.java
+++ b/src/amitaRJun20/CheckCharacterMethods.java
@@ -1,6 +1,7 @@
 package amitaRJun20;
 
 public class CheckCharacterMethods {
+	
 	void testCharMethod(String name,char ch) {
 		int digitCount=0;
 		int letterCount=0;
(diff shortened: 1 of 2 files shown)


src/amitaRJun20/CheckCharacterMethods.java
```diff
@@ -0,0 +1,39 @@
...
+				digitCount++;
+			}
+			if(Character.isLetter(name.charAt(index))) {
+				letterCount++;
+			}
+			if(Character.isUpperCase(name.charAt(index))) {
+				upperCaseCount++;
+			}
+			if(Character.isLowerCase(name.charAt(index))) {
+				lowerCaseCount++;
```
KrishnaKTechnocredits: Here so many comparisons happening, which is not required. 
Second point letterCount without using this variable you can print the expected output. Think about it.

src/amitaRJun20/FindIndexOfCharacters.java
```diff
@@ -0,0 +1,21 @@
...
+public class FindIndexOfCharacters {
+	
+	void findIndexOfChar() {
+		
+		String name="technocredits";
+		System.out.println("The given String is : "+name+"\n");		
+		for(int index=0;index<name.length();index++) {
+			char ch=name.charAt(index);
+			if(index==name.indexOf(ch) ) {
+				if(name.indexOf(ch) != name.lastIndexOf(ch))
```
KrishnaKTechnocredits, line 12: Perfect Logic.

src/amitaRJun20/CheckCharacterMethods.java
AmitaRJun20: Thank you for your feedback....
after i committed the code then i realized after reading your comment few times that letter count variable not required....so again updated the logic and committed...sorry for multiple commits...

src/amitaRJun20/CheckCharacterMethods.java
```diff
@@ -0,0 +1,38 @@
...
+					upperCaseCount++;
+				else 
+					lowerCaseCount++;
+				}			
+			else 
+				specialCharCount++;
+		}
+		System.out.println("The given string is "+name);
+		System.out.println("total Digit : "+digitCount);
+		System.out.println("Total Letters : "+(upperCaseCount+lowerCaseCount));
```
KrishnaKTechnocredits, line 24: Here it's correct.

src/amitaRJun20/CheckCharacterMethods.java
```diff
@@ -0,0 +1,38 @@
...
+		int digitCount=0;
+		int upperCaseCount=0;
+		int lowerCaseCount=0;
+		int specialCharCount=0;
+		
+		for(int index=0;index<name.length();index++) {
+			if(Character.isDigit(name.charAt(index))) {
+				digitCount++;
+			}
+			else if(Character.isLetter(name.charAt(index))) {
```
KrishnaKTechnocredits: Here else if(Character.isLetter(name.charAt(index))) you don't need this line beacuse letters will already going to cover under uppercase and lowercase. Just remove this line and write an individual else if block for uppercase and lower case.

src/amitaRJun20/CheckCharacterMethods.java
KrishnaKTechnocredits: No need to say sorry, it's your PR...you can make as many commits you want.

PR #435, opened 2020-06-25
Harshad_Assignment_17_Done_50mins

HarshadBJun20:
->Assignment 17 Done.
-> Screenshot Program 1 of output:
image

->Screenshot Program 2 of output:
image

### DIFF
--- a/src/harshadBJun20/CountOfCharsUsingWrapper.java
+++ b/src/harshadBJun20/CountOfCharsUsingWrapper.java
@@ -13,24 +13,22 @@ Total Lowercase: 3
 Total Special char: 2*/
 
 public class CountOfCharsUsingWrapper{
-	int digitCount,letterCount,uppercaseCount,lowercaseCount,specialCharCount;
+	int digitCount,uppercaseCount,lowercaseCount,specialCharCount;
 	void printCountOfChars(String str1) {
 		
 		for(int index=0;index<str1.length();index++){
+			
 			if(Character.isDigit(str1.charAt(index))==true)
-				digitCount++;
-			else if(Character.isLetter(str1.charAt(index))==true)
-				letterCount++;
-			else
-				specialCharCount++;
+				digitCount++;  			
 			if(Character.isUpperCase(str1.charAt(index))==true)
 				uppercaseCount++;
 			if(Character.isLowerCase(str1.charAt(index))==true)
 				lowercaseCount++;	
+			specialCharCount = (str1.length() - digitCount - uppercaseCount - lowercaseCount);
 			
 		}
 		System.out.println("Total digit: "+digitCount);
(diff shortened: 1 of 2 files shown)


src/harshadBJun20/CountOfCharsUsingWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class CountOfCharsUsingWrapper{
+	int digitCount,letterCount,uppercaseCount,lowercaseCount,specialCharCount;
+	void printCountOfChars(String str1) {
+		
+		for(int index=0;index<str1.length();index++){
+			if(Character.isDigit(str1.charAt(index))==true)
+				digitCount++;
+			else if(Character.isLetter(str1.charAt(index))==true)
+				letterCount++;
```
KrishnaKTechnocredits: letterCount without using this variable and else if block you can achieve expected output. Think about it.

src/harshadBJun20/FirstLastIndex.java
```diff
@@ -0,0 +1,40 @@
...
+	void returnFirstIndex(String str){
+		
+		for(int index=0;index<str.length();index++){
+			char ch = str.charAt(index);
+			if(index == str.indexOf(ch) && index!=str.lastIndexOf(ch)){
+				System.out.print(ch+" -> first index = "+index);
+				//if(index == str.indexOf(ch)) {
+					for(int j=0;j<str.length();j++){
+						if(j==str.lastIndexOf(ch)){
+							System.out.println(" last index = "+j);
```
KrishnaKTechnocredits: You have increased the complexity in code, Using one for loop and one if block you can achieve the output.

HarshadBJun20: -> Changes done as per comments. Please review. Thanks.

src/harshadBJun20/CountOfCharsUsingWrapper.java
HarshadBJun20: -> Changes done as per comments. Please review. Thanks.

src/harshadBJun20/CountOfCharsUsingWrapper.java
```diff
@@ -0,0 +1,45 @@
...
+	void printCountOfChars(String str1) {
+		
+		for(int index=0;index<str1.length();index++){
+			
+			if(Character.isDigit(str1.charAt(index))==true)
+				digitCount++;  			
+			if(Character.isUpperCase(str1.charAt(index))==true)
+				uppercaseCount++;
+			if(Character.isLowerCase(str1.charAt(index))==true)
+				lowercaseCount++;
```
KrishnaKTechnocredits, line 26: Here so many comparisons happening, which is not required. 

KrishnaKTechnocredits (approved):

PR #436, opened 2020-06-25
Yash_Assignment_17_90min_Done

YashCJun20:
Output : Program 1
image

Output :Program 2
image

### DIFF
--- a/src/yashCJun20/CharOccurance.java
+++ b/src/yashCJun20/CharOccurance.java
@@ -8,18 +8,15 @@ public class CharOccurance {
 		{
 		char ch=name.charAt(i);
 		if(i == name.indexOf(ch) && i!=name.lastIndexOf(ch)){
-			System.out.print(ch+" -> first index = "+i);
-		for(int index=0;index<name.length();index++) {
-			if(index==name.lastIndexOf(ch)){
-				System.out.println(" last index = "+index);
+			System.out.print(ch+" -> first index = "+name.indexOf(ch));
+			System.out.println(" Last index = "+name.lastIndexOf(ch));
+			
 			}
 		}
 		
 		}
-		}
 		
-	}
-
+	
 	public static void main(String[] args) {
 		
 		CharOccurance charOccurance = new CharOccurance();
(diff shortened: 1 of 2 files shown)


src/yashCJun20/CharOccurance.java
```diff
@@ -0,0 +1,32 @@
...
+	void Charindex(String name) {
+		
+		for(int i=0;i<name.length();i++)
+		{
+		char ch=name.charAt(i);
+		if(i == name.indexOf(ch) && i!=name.lastIndexOf(ch)){
+			System.out.print(ch+" -> first index = "+i);
+		for(int index=0;index<name.length();index++) {
+			if(index==name.lastIndexOf(ch)){
+				System.out.println(" last index = "+index);
```
KrishnaKTechnocredits: You have increased the complexity in code, Using one for loop and one if block you can achieve the output.

src/yashCJun20/Count.java
```diff
@@ -0,0 +1,47 @@
...
+		int lowerCaseCnt=0;
+		int specialCharCnt=0;
+		
+		for (int i=0;i<s.length();i++) {
+			char ch = s.charAt(i);
+			
+			if(Character.isDigit(ch)==true)
+				digitCnt++;
+			else if(Character.isLetter(ch)==true)
+				letterCnt++;
```
KrishnaKTechnocredits: letterCnt without using this variable and else if block you can achieve expected output.

src/yashCJun20/Count.java
```diff
@@ -0,0 +1,47 @@
...
+			char ch = s.charAt(i);
+			
+			if(Character.isDigit(ch)==true)
+				digitCnt++;
+			
+			if(Character.isUpperCase(ch)==true)
+				upperCaseCnt++;
+			
+			if(Character.isLowerCase(ch)==true)
+				lowerCaseCnt++;
```
KrishnaKTechnocredits, line 23: Here so many comparisons happening, which is not required. 

KrishnaKTechnocredits (approved):

PR #442, opened 2020-06-25
Vaishnavi_Assignment_17_35Min_Done

VaishnaviVJun20:
image
image

### DIFF
--- a/src/vaishnaviVJun20_arrayExamples/IndexOfEx.java
+++ b/src/vaishnaviVJun20_arrayExamples/IndexOfEx.java
@@ -3,17 +3,11 @@ package vaishnaviVJun20_arrayExamples;
 public class IndexOfEx {
 	
 	void printFirstAndLastIndex(String s) {
+		
 		for(int index=0;index<s.length();index++) {
 			char ch = s.charAt(index);
-			int count=0;
-			if(index == s.indexOf(ch)) {
-				for(int i=0;i<s.length();i++) {
-					if(ch== s.charAt(i))
-						count++;				
-				}
-				if(count>1)
-					System.out.println(ch +"--> first index= " +s.indexOf(ch) +" and Last index=" +s.lastIndexOf(ch) );
-			}
+			if(index == s.indexOf(ch) && s.indexOf(ch) != s.lastIndexOf(ch))
+				System.out.println(ch +"--> first index= " +s.indexOf(ch) +" and Last index=" +s.lastIndexOf(ch) );
 		}
 	}
 
--- a/src/vaishnaviVJun20_arrayExamples/IndexOfEx.java
+++ b/src/vaishnaviVJun20_arrayExamples/IndexOfEx.java
@@ -1,0 +1,26 @@
+package vaishnaviVJun20_arrayExamples;
+
+public class IndexOfEx {
+	
+	void printFirstAndLastIndex(String s) {
+		for(int index=0;index<s.length();index++) {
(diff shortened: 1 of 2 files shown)


src/vaishnaviVJun20_arrayExamples/IndexOfEx.java
```diff
@@ -0,0 +1,26 @@
...
+	
+	void printFirstAndLastIndex(String s) {
+		for(int index=0;index<s.length();index++) {
+			char ch = s.charAt(index);
+			int count=0;
+			if(index == s.indexOf(ch)) {
+				for(int i=0;i<s.length();i++) {
+					if(ch== s.charAt(i))
+						count++;				
+				}
```
KrishnaKTechnocredits: You have increased the complexity in code, Using one for loop and one if block you can achieve the output.

src/vaishnaviVJun20_arrayExamples/StringMethods.java
```diff
@@ -0,0 +1,36 @@
...
+		int letterCount=0;
+		int specialCharCount=0;
+		for(int index=0;index<s.length();index++) {
+			if(Character.isDigit(s.charAt(index)))
+				digitCount++;
+			if(Character.isLetter(s.charAt(index)))
+				letterCount++;
+			if(Character.isUpperCase(s.charAt(index)))
+				upperCaseCount++;
+			if(Character.isLowerCase(s.charAt(index)))
```
KrishnaKTechnocredits: Here so many comparisons happening, which is not required. 
Second point letterCount without using this variable you can print the expected output. Think about it.

src/vaishnaviVJun20_arrayExamples/IndexOfEx.java
VaishnaviVJun20: Thank you for suggestion, Changes done.

src/vaishnaviVJun20_arrayExamples/StringMethods.java
VaishnaviVJun20: Changes done, I wasn't thinking about that letter logic before that's why so many if's. Thanks :)

src/vaishnaviVJun20_arrayExamples/IndexOfEx.java
```diff
@@ -0,0 +1,20 @@
+package vaishnaviVJun20_arrayExamples;
+
+public class IndexOfEx {
+	
+	void printFirstAndLastIndex(String s) {
+		
+		for(int index=0;index<s.length();index++) {
+			char ch = s.charAt(index);
+			if(index == s.indexOf(ch) && s.indexOf(ch) != s.lastIndexOf(ch))
```
KrishnaKTechnocredits, line 9: Perfect logic.

src/vaishnaviVJun20_arrayExamples/StringMethods.java
```diff
@@ -0,0 +1,33 @@
...
+				digitCount++;
+			else if(Character.isUpperCase(s.charAt(index)))
+				upperCaseCount++;
+			else if(Character.isLowerCase(s.charAt(index)))
+				lowerCaseCount++;
+			else
+				specialCharCount++;
+		}
+		System.out.println("Total Digit: "+digitCount);
+		System.out.println("Total Letter: "+(lowerCaseCount+upperCaseCount));
```
KrishnaKTechnocredits, line 21: Good job Amita, keep it up.

PR #443, opened 2020-06-25
Abhijit_Assignment_17_50min_Done

AbhijitMJun20:
WhatsApp Image 2020-06-25 at 10 51 07 PM
WhatsApp Image 2020-06-25 at 10 51 07 PM (1)

### DIFF
--- a/src/abhijitMJun20/CountLetters.java
+++ b/src/abhijitMJun20/CountLetters.java
@@ -1,0 +1,31 @@
+package abhijitMJun20;
+
+public class CountLetters {
+	void charfind(String input)
+	{ 
+		 int count=0,count1=0,count2=0,count3=0;
+		for(int index=0;index<input.length();index++) {
+		if(Character.isUpperCase(input.charAt(index))){
+		count++;
+		}
+		else if(Character.isLowerCase(input.charAt(index))) {
+			count1++;
+		}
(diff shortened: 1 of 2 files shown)


src/abhijitMJun20/CountLetters.java
```diff
@@ -0,0 +1,31 @@
+package abhijitMJun20;
+
+public class CountLetters {
+	void charfind(String input)
+	{ 
+		 int count=0,count1=0,count2=0,count3=0;
```
KrishnaKTechnocredits, line 6: First thing I really don't like the way you have given variable name. Give a proper name.
Like countLowercase...etc. Update it after I will check the rest of the code.

src/abhijitMJun20/Indexofuse.java
```diff
@@ -0,0 +1,35 @@
+package abhijitMJun20;
+
+public class Indexofuse {
+	void indexuseof(String name){
+		for(int index=0;index<name.length();index++){
+			char ch= name.charAt(index);
+			if(index==name.indexOf(ch)) {
```
KrishnaKTechnocredits, line 7: Here you should add one more if a condition that is (name.indexOf(ch) != name.lastIndexOf(ch))

PR #445, opened 2020-06-25
Archana_Assigment_17_IndexOfExample1_15min_IndexOfExample2_30min_Done

ArchanaNJun20:
IndexOfExample1
image

IndexOfExample2
image

### DIFF
--- a/src/archanaNJun20/IndexOfExample1.java
+++ b/src/archanaNJun20/IndexOfExample1.java
@@ -1,0 +1,39 @@
+package archanaNJun20;
+
+public class IndexOfExample1 {
+
+	void countStringChar(String input)
+	{
+		int digitCount=0;
+		int uCount=0;
+		int lCount=0;
+		int loCount=0;
+		int spCount=0;
+
+		for(int index=0;index<input.length();index++){
+			char ch=input.charAt(index);
+			if(Character.isDigit(ch)) 
+				digitCount++;
+			else  if(Character.isLetter(ch)) {
+				lCount++;
+				if(Character.isUpperCase(ch)) 
+					uCount++;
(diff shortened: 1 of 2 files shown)


src/archanaNJun20/IndexOfExample1.java
```diff
@@ -0,0 +1,39 @@
...
+		int loCount=0;
+		int spCount=0;
+
+		for(int index=0;index<input.length();index++){
+			char ch=input.charAt(index);
+			if(Character.isDigit(ch)) 
+				digitCount++;
+			else  if(Character.isLetter(ch)) {
+				lCount++;
+				if(Character.isUpperCase(ch))
```
KrishnaKTechnocredits, line 16: Without lCount use of this variable and else  if(Character.isLetter(ch))  block you are able to achieve the output.

src/archanaNJun20/IndexOfExample2.java
```diff
@@ -0,0 +1,26 @@
...
+		for(int index=0;index<input.length();index++){
+			char ch=input.charAt(index);
+			int count=0;
+			if(index==input.indexOf(ch))
+				for(int index1=0;index1<input.length();index1++) {
+					if(ch==input.charAt(index1)){
+						count++;
+					}
+				}
+			if(count>1){
```
KrishnaKTechnocredits: You have increased the complexity in code, Using one for loop and one if block you can achieve the output.

src/archanaNJun20/IndexOfExample1.java
KrishnaKTechnocredits: Here so many comparisons happening, which is not required. 

PR #451, opened 2020-06-26
Kartikey_Assignment_17_Done_45min

KartikeyDJune20:
image
image

### DIFF
--- a/src/kartikeyDJune20/CountElementsinPreDefindString.java
+++ b/src/kartikeyDJune20/CountElementsinPreDefindString.java
@@ -1,0 +1,37 @@
+package kartikeyDJune20;
+
+public class CountElementsinPreDefindString {
+	static void countInString(String input) {
+		int cntDig = 0;
+		int cntUpper = 0;
+		int cntLower = 0;
+		int cntSpecial = 0;
+		for (int index = 0; index < input.length(); index++) {
+
+			char ch = input.charAt(index);
+			if (Character.isDigit(ch))
+				cntDig++;
+			else if (Character.isLetter(ch)) {
+
+					if (Character.isLowerCase(ch))
+					cntLower++;
+					else
+					cntUpper++;
+			} else
+				cntSpecial++;
+
+		}
+		System.out.println("There are " + cntDig + " digits in " + input);
+		System.out.println("There are " + cntLower + " lower in " + input);
+		System.out.println("There are " + cntUpper + " Uppercase in " + input);
+		System.out.println("There are " + cntSpecial + " Special Characters in " + input);
+		System.out.println("There are "+(cntUpper+cntLower)+" Letter in "+input);
+	}
+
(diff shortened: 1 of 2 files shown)


src/kartikeyDJune20/CountElementsinPreDefindString.java
```diff
@@ -0,0 +1,37 @@
...
+		int cntSpecial = 0;
+		for (int index = 0; index < input.length(); index++) {
+
+			char ch = input.charAt(index);
+			if (Character.isDigit(ch))
+				cntDig++;
+			else if (Character.isLetter(ch)) {
+
+					if (Character.isLowerCase(ch))
+					cntLower++;
```
KrishnaKTechnocredits, line 17: else if (Character.isLetter(ch)) without this block you are able to achieve the output.

src/kartikeyDJune20/CountElementsinPreDefindString.java
```diff
@@ -0,0 +1,37 @@
...
+					cntUpper++;
+			} else
+				cntSpecial++;
+
+		}
+		System.out.println("There are " + cntDig + " digits in " + input);
+		System.out.println("There are " + cntLower + " lower in " + input);
+		System.out.println("There are " + cntUpper + " Uppercase in " + input);
+		System.out.println("There are " + cntSpecial + " Special Characters in " + input);
+		System.out.println("There are "+(cntUpper+cntLower)+" Letter in "+input);
```
KrishnaKTechnocredits, line 28: This logic is correct for letter.

src/kartikeyDJune20/FirstLastIndexInString.java
```diff
@@ -0,0 +1,24 @@
...
+	
+	static void firstLastIndexInString(String input) {
+		
+		for(int index=0;index< input.length();index++) {
+			int firstIndex=0;
+			int lastIndex=0;
+			char ch=input.charAt(index);
+			firstIndex=input.indexOf(ch);
+			lastIndex=input.lastIndexOf(ch);
+			if(firstIndex!=index && lastIndex!=0)
```
KrishnaKTechnocredits, line 13: Are you sure you need so many local variable for this, no!!!
for (int index = 0; index < name.length(); index++) {
			char ch = name.charAt(index);
			if (index == name.indexOf(ch) && name.indexOf(ch) < name.lastIndexOf(ch)) 
					System.out.println(
							ch + " -> first index = " + name.indexOf(ch) + " Last index = " + name.lastIndexOf(ch));
		}

src/kartikeyDJune20/CountElementsinPreDefindString.java
KrishnaKTechnocredits: No need to check for letters, because it will already going to cover under uppercase and lowercase.

PR #452, opened 2020-06-26
AnupS_Assignment_17_Done_30mins

AnupSJun20:
AnupS_Assignment_17_Done. Pl
Program1
Program2

ease review.

### DIFF
--- a/src/anupSJun20/CharacterMethods.java
+++ b/src/anupSJun20/CharacterMethods.java
@@ -1,0 +1,39 @@
+package anupSJun20;
+/*1) Count Digits,Letters, UpperCase characters, LowerCase characters, Special characters from predefined string.
+
+Input: 1rRpd3F9#K(E
+Output : 
+Total digit: 3
+Total Letters: 7
+Total Uppercase: 4
+Total Lowercase: 3
+Total Special char: 2*/
+
+public class CharacterMethods {
+	static void findCharCountInString(String str) {
+		int digitCount = 0;
+		int uppercase = 0;
+		int lowercase = 0;
+		int specialChar = 0;
+		for(int index = 0;index<str.length();index++) {
+			char ch = str.charAt(index);
+			if (Character.isDigit(ch)==true)
+				digitCount++;
+			else if(Character.isLetter(ch) == true) {
+			if(Character.isUpperCase(ch) == true)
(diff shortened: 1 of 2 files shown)


src/anupSJun20/CharacterMethods.java
```diff
@@ -0,0 +1,39 @@
...
+		int digitCount = 0;
+		int uppercase = 0;
+		int lowercase = 0;
+		int specialChar = 0;
+		for(int index = 0;index<str.length();index++) {
+			char ch = str.charAt(index);
+			if (Character.isDigit(ch)==true)
+				digitCount++;
+			else if(Character.isLetter(ch) == true) {
+			if(Character.isUpperCase(ch) == true)
```
KrishnaKTechnocredits: First thing why are you  here writing == true, because if a condition will be true then only it will go in the respective block. 
Second thing else if(Character.isLetter(ch) == true) this if block is not needed. 
Because it will already going to cover under uppercase and lowercase.

src/anupSJun20/CharacterMethods.java
```diff
@@ -0,0 +1,39 @@
...
+				uppercase++;
+			else
+				lowercase++;
+			}
+			else
+				specialChar++;
+		}
+		System.out.println(" Total digit : "+ digitCount + "\n Total Letters: " + (uppercase+lowercase) + "\n Total Uppercase: "+ uppercase);
```
KrishnaKTechnocredits, line 29: Here for letter you have written correct logic.

src/anupSJun20/MultipleCharacters.java
```diff
@@ -0,0 +1,28 @@
...
+c -> first index = 2 Last index = 6
+
+Hint : To find first index of character --> indexOf() method
+            To find last index of character --> lastIndexOf() method.*/
+
+public class MultipleCharacters {
+	static void findMultipleCharacters(String str) {
+		for(int index = 0; index<str.length();index++) {
+			char ch = str.charAt(index);
+			if(index != str.lastIndexOf(ch)) {
```
KrishnaKTechnocredits, line 17: Perfect logic.

KrishnaKTechnocredits (approved):
Now it's perfect.

PR #463, opened 2020-06-26
Pranita assignment 18 Done _Total 2hours

PranitaPJun20:
freqOfword
middlechar
reverse string
totalSpclchar


KrishnaKTechnocredits (approved):
Code looks good to me.

PR #464, opened 2020-06-26
Barkha assignment 18

BarkhaPJun20:
Program 1 :
Find Frequency of every word in a given String.
Rules:
a) Word should be compare without case sensitivity.
b) Word should be compare without special character. [use replace method]

Input = "Tech_no Hi T_EchNo H_I Tech hi_"
output : 
Frequency of every word:
              techno -> 2
              hi -> 3
              tech -> 1
Program 2 : 
Create a new method to print middle character of all words.
Input = "Harsh Maulik";
Output : r is middle character in Harsh
              u is middle character in Maulik
Program 3 : 
Count total special characters like _ , [ { ] } from every word, do not use any loop.
Hint : Use replace method [Think about it, how it can help].
Input :  Kr[i{s]hn}_a,Tech
Output : total special characters are 6 
 Program 4 : 
Print given string in reverse order.
Input : Hello this is technocredits
Outtput : stiderconhcet si siht olleH

image
image
image
image

### DIFF
--- a/src/barkhaPJun20/ClassTestDebug.java
+++ b/src/barkhaPJun20/ClassTestDebug.java
@@ -1,0 +1,26 @@
+package barkhaPJun20;
+
+public class ClassTestDebug {
+
+	void frequencyOfNumber(int[] array, int number) {
+		int count = 0;
+
+		for (int index = 0; index < array.length; index++) {
+			if (array[index] == number)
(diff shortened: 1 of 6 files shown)


KrishnaKTechnocredits (approved):
Code looks good to me.

PR #467, opened 2020-06-26
ADT_Assignment18_Done_(*+10+5+5)Mins

AditiGJUN20:
**Assignment 1:**
- WordFreqInString.java
Time: *Lost track of time
Screen Shot 2020-06-26 at 11 02 32 AM
- Commented the method which Maulik had explained in session as reference
- I'd used length and replace while i'd written code - So, kept it and made the tweaks as Maulik explained.

**Assignment 2:**
- ExtractMiddleChar.java
- Time: 10 Mins
Screen Shot 2020-06-26 at 10 27 52 AM

**Assignment 3:**
- CountSpecialChar.java
- Time: 5 Mins
Screen Shot 2020-06-26 at 10 29 38 AM

**Assignment 4:**
- ReverseString.java
- Time: 5Mins
Screen Shot 2020-06-26 at 10 26 54 AM

### DIFF
--- a/src/aditiGJUN20/ExtractMiddleChar.java
+++ b/src/aditiGJUN20/ExtractMiddleChar.java
@@ -1,0 +1,32 @@
+/*
+Create a new method to print middle character of all words.
+
+Input = "Harsh Maulik";
+Output : r is middle character in Harsh
+              u is middle character in Maulik 
+ 
+*/
+
+package aditiGJUN20;
+
+public class ExtractMiddleChar {
+	
+	// Method to find middle character of String.
+	void getMidCharOfString(String[] array) {
+		for(int i=0;i<array.length;i++) {
+			String name=array[i];
+			if(name.length()%2 ==0)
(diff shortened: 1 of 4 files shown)


src/aditiGJUN20/ExtractMiddleChar.java
```diff
@@ -0,0 +1,32 @@
...
+package aditiGJUN20;
+
+public class ExtractMiddleChar {
+	
+	// Method to find middle character of String.
+	void getMidCharOfString(String[] array) {
+		for(int i=0;i<array.length;i++) {
+			String name=array[i];
+			if(name.length()%2 ==0)
+				System.out.println(String.valueOf(name.charAt((name.length()/2)-1))+" is middle character in "+name);
```
KrishnaKTechnocredits, line 19: This will surely work.. but what about below code.
For even or odd length string.
(name.length()-1)/2. This would be without condition.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #474, opened 2020-06-27
AMITA_Assignment_18_Done_90mins

AmitaRJun20:
AMITA_Assignment_18_Done_90mins....Please Review...
countSpecialChar
MiddleChar
ReverseString
wordFrequency

### DIFF
--- a/src/amitaRJun20/PrintMiddleCharOfWord.java
+++ b/src/amitaRJun20/PrintMiddleCharOfWord.java
@@ -1,0 +1,24 @@
+package amitaRJun20;
+
+public class PrintMiddleCharOfWord {
+	
+	void findMiddleChar() {
+		String input="Harsh Maulik";
+		System.out.println("print middle character of : "+input);
+		String[] array=input.split(" ");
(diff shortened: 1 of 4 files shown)


src/amitaRJun20/PrintMiddleCharOfWord.java
```diff
@@ -0,0 +1,24 @@
...
+public class PrintMiddleCharOfWord {
+	
+	void findMiddleChar() {
+		String input="Harsh Maulik";
+		System.out.println("print middle character of : "+input);
+		String[] array=input.split(" ");
+		for(int index=0;index<array.length;index++) {
+			String name=array[index];
+			int middleChar=(name.length())/2;
+			if(name.length()%2==0)
```
KrishnaKTechnocredits, line 12: without condition also we can do it.
Think about it.
Harsh --> 2
Maulik --> 2

what if we say name.charAt(name.length()-1)

KrishnaKTechnocredits (approved):
Code looks good to me. Please take a look at comments.

PR #475, opened 2020-06-27
Rachana_Assignment_18_90min_Done

RachanaGJun20:
prg1
Middlechar
prg3
prg4

### DIFF
--- a/src/rachanaGJun20/String/StringReplaceFun.java
+++ b/src/rachanaGJun20/String/StringReplaceFun.java
@@ -1,0 +1,34 @@
+/*Count total special characters like _ , [ { ] } from every word, do not use any loop.
+Hint : Use replace method [Think about it, how it can help].
+
+Input :  Kr[i{s]hn}_a,Tech
+Output : total special characters are 6 */
+package rachanaGJun20.String;
+
+public class StringReplaceFun {
+
+	// method will print special char count in a given string
+	void countSpecialChar(String givenstr) {
+		
+
+		String newstr = givenstr;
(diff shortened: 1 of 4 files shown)


KrishnaKTechnocredits (approved):
Very good job. Code looks good to me.

src/rachanaGJun20/String/StringReplaceFun.java
```diff
@@ -0,0 +1,34 @@
...
+package rachanaGJun20.String;
+
+public class StringReplaceFun {
+
+	// method will print special char count in a given string
+	void countSpecialChar(String givenstr) {
+		
+
+		String newstr = givenstr;
+		newstr = newstr.replace("Kr", "");
```
KrishnaKTechnocredits, line 15: We should not write it like this. This is completely hardcoded... Your code wont work if name is maulik with same special character. and givenstr has no meaning then why to pass parameter if you want to say replace Kr with "" and so on...

RachanaGJun20: > Very good job. Code looks good to me.

Thanks :)

RachanaGJun20: > We should not write it like this. This is completely hardcoded... Your code wont work if name is maulik with same special character. and givenstr has no meaning then why to pass parameter if you want to say replace Kr with "" and so on...

yeah right got it.I have changed method calling , but not able to write other logic(without hardcoading) with out using loop. please suggest.

PR #480, opened 2020-06-27
Pranita_Coding_Exam_2_Done_10mins

PranitaPJun20:
image


harshvegada (approved):
Logic is correct so far
but no need to define extra variables like countDigit, lilttle bit code improvement needed in "if" condition for temp variable. variable name should be proper (i.e. here temp make no sense to use)
Do not print extra stuff like countDigit on 31st line

PR #484, opened 2020-06-27
Deavina_Coding_Exam_2_Done_6mins

DeavinaSJun20:
image

### DIFF
--- a/src/deavinaSJun20/TestIntSumClass.java
+++ b/src/deavinaSJun20/TestIntSumClass.java
@@ -1,4 +1,5 @@
 package deavinaSJun20;
+
 /*String input = "1Hh9PR34QP";
 Hint : 1+9+3+4 [all digits]
 output : 178*/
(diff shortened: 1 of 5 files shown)


src/deavinaSJun20/TestIntSumClass.java
```diff
@@ -0,0 +1,25 @@
...
+output : 178*/
+public class TestIntSumClass {
+
+	public static void main(String[] args) {
+		// TODO Auto-generated method stub
+String name="1Hh9PR34QP";
+
+new TestIntSumClass().intSumClass(name);
+	}
+  void intSumClass(String name) {
```
harshvegada: Logic is correct.
Give proper method name so third can understand by reading method name.
rest is good
Keep it up

DeavinaSJun20: updated file checked in as per review comments . please review

harshvegada: Okay merging PR

PR #489, opened 2020-06-27
Barkha coding exam 2_15 Mins

BarkhaPJun20:
String input = "1Hh9PR34QP";
Hint : 1+9+3+4 [all digits]
output : 17
To convert char to Integer.
Points to be remember:
1) String can be converted to int but there is no direct method to convert char to int
2) char has to be coverted first to string and then to interger.
image

### DIFF
--- a/src/barkhaPJun20/CountSpecialCharUsingReplaceMethod.java
+++ b/src/barkhaPJun20/CountSpecialCharUsingReplaceMethod.java
@@ -1,0 +1,27 @@
+package barkhaPJun20;
+
+public class CountSpecialCharUsingReplaceMethod {
+	
+	void findSpecialChar(String name) {
+		
+		String name1=name.replace("{", "").replace("[", "").replace("]", "").replace(",", "").replace("}", "");						
+				
+			//String name1=name.replace("{", "");
+			//String name2=name1.replace("[", "");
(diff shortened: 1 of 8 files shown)


src/barkhaPJun20/SumOfDigitsInString.java
```diff
@@ -0,0 +1,31 @@
+package barkhaPJun20;
+
+public class SumOfDigitsInString {
+	
+	
+	void addition(String name) {
```
harshvegada: @BarkhaPJun20 
Code logic up the mark, keep it up.
Give method name proper way. (work on this)

BarkhaPJun20: Updated proper method name.

src/barkhaPJun20/CountSpecialCharUsingReplaceMethod.java
```diff
@@ -0,0 +1,27 @@
+package barkhaPJun20;
+
+public class CountSpecialCharUsingReplaceMethod {
+	
+	void findSpecialChar(String name) {
+
```
KrishnaKTechnocredits, line 6: Remove extra blank line.

PR #490, opened 2020-06-27
Madhura_Coding_Exam_2_18mins

MadhuraMJun20:
image


harshvegada (approved):
Code logic is correct, one improvement point 
store this in one variable : input.charAt(index) and then use this multiple way, if you check line number 11(you are using this method to retrieve char) and same method you are using to get char on line number 12

this is something which you need to improve, rest is good

PR #492, opened 2020-06-27
Kartikey_Coding_Exam_2_Done_15mins

KartikeyDJune20:
image


harshvegada (approved):
Code logic is correct, but do not go for "+" to get String value, use
String.valueOf(char) method to get char into String
Do not print anything inside the method

Rest code logic is correct.

PR #493, opened 2020-06-27
Amita_Coding_Exam_2_Done_10mins

AmitaRJun20:
Amita_Coding_Exam_2_Done_10mins...please review...
Capture


harshvegada (approved):
Code is logic is up to the mark @AmitaRJun20 
Keep it up

Give method name proper so third person can understand method use.
Rest is good!

PR #499, opened 2020-06-27
Karan code exam 30mins

KARAN1595:
completed


harshvegada (approved):
Logic is correct!

Give variable name properly so, third person can understand use of by reading variable name.

PR #502, opened 2020-06-27
Anshu-CodeExam-2-30min

AnshuKJun20:


harshvegada (approved):
Coding logic is correct @AnshuKJun20 



Give method/variable name properly so, third person can understand use of by reading method name.
Here we not checking prime number. we are doing addition of digit which is in String format.

remove extra coding if not required (commented code)

PR #506, opened 2020-06-27
Viresh_Assignment_19_done_(20+15+5+5+10)Min

VireshJJun20:
Please review and comment
image

image

image

image

image

### DIFF
--- a/src/vireshJJun20/Assignment19/ArrayComparison3.java
+++ b/src/vireshJJun20/Assignment19/ArrayComparison3.java
@@ -1,0 +1,28 @@
+package vireshJJun20.Assignment19;
+
+public class ArrayComparison3 { //// returns the maximum value from mismatching index post comparing 2 arrays.
+	int index= 0;
+	void arrayIndex(int[] ar1, int[] ar2) {
+		if (ar1.length==ar2.length) {
+			for (int i= 0; i<ar2.length; i++) {
+				if(ar1[i]!=ar2[i]) {
+					int max=0;
+					if (ar1[i]> ar2[i]) {
+						max= ar1[i];
+					}	
+					else {
+						max= ar2[i];
+					}
+						System.out.println("Values are not matching at index " + i + ", from (" + ar1[i] + "," + ar2[i] + ") max value is " + max);
+				}
+			}
+		}	
+	}
+	
+	public static void main(String[] args) {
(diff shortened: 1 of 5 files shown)


src/vireshJJun20/Assignment19/ArrayComparison3.java
```diff
@@ -0,0 +1,28 @@
+package vireshJJun20.Assignment19;
+
+public class ArrayComparison3 { //// returns the maximum value from mismatching index post comparing 2 arrays.
+	int index= 0;
+	void arrayIndex(int[] ar1, int[] ar2) {
+		if (ar1.length==ar2.length) {
+			for (int i= 0; i<ar2.length; i++) {
+				if(ar1[i]!=ar2[i]) {
+					int max=0;
```
KrishnaKTechnocredits, line 9: you should define max = ar1[2] initially, else is not required.

src/vireshJJun20/Assignment19/OpsPrime.java
```diff
@@ -0,0 +1,64 @@
...
+		System.out.println("Average of all " + count + " prime numbers in between 2- 100 is " + avg);
+		
+	}
+
+	public static void main(String[] args) {
+		OpsPrime opsPrime = new OpsPrime();
+		int[] ar= {7, 9, 13};
+		System.out.println("Given array: {7, 9, 13}");
+		for (int i=0; i<ar.length;i++) { //  selecting a single array element and sending it to findPrime method to check if its a prime number or not. displays a message based on the return
+			int num= ar[i];
```
KrishnaKTechnocredits, line 49: for loop code should be written in a separate method instead of main method.

src/vireshJJun20/Assignment19/ArrayComparison3.java
VireshJJun20: Ok Krishna, will take care from next time. Thanks.

src/vireshJJun20/Assignment19/OpsPrime.java
VireshJJun20: Agreed Krishna, but the reason I had added a loop in main is to send one by one char to a predefined method which returns if number is prime or no.
I would write a new method with the specified loop, from next time. 
Thank you.

PR #507, opened 2020-06-27
Aavruti_Assignment_19_(27+5+5+5+5)mins

AavrutiDJun20:
MaxValue
PrimeNo
StringArrayDiffer
AllDiffer
DifferArray

### DIFF
--- a/src/aavrutiDJun20/PrimeNumbers.java
+++ b/src/aavrutiDJun20/PrimeNumbers.java
@@ -1,0 +1,45 @@
+package aavrutiDJun20;
+
+public class PrimeNumbers {
+	
+	//Print whether given number is Prime Number or not
+	void primeNumberOrNot(int number) {
+		boolean flag = false;
+		for(int index=2;index<number/2;index++) {
+			if(number%index==0)
+				flag = true;
+		}
+		if(flag==true)
+			System.out.println(number + " is not a Prime Number");
+		else
+			System.out.println(number + " is a Prime Number");		
+	}
+	
+	//Calculate Sum, Count and Average of Prime Numbers
+	void primeNumber2To100() {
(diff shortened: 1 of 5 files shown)


src/aavrutiDJun20/PrimeNumbers.java
```diff
@@ -0,0 +1,47 @@
...
+				flag = true;
+		}
+		if(flag==true)
+			System.out.println(number + " is not a Prime Number");
+		else
+			System.out.println(number + " is a Prime Number");		
+	}
+	
+	//Calculate Sum, Count and Average of Prime Numbers between 2 - 100
+	void primeNumber2To100() {
```
KrishnaKTechnocredits: @AavrutiDJun20 it is better to take parameter, like startNum and endNum , so that tomorrow if range change to 2 to 200 you don't need to update your loop.

AavrutiDJun20: Done

src/aavrutiDJun20/PrimeNumbers.java
```diff
@@ -0,0 +1,49 @@
...
+			}
+		}
+		if(flag==true)
+			System.out.println(number + " is not a Prime Number");
+		else
+			System.out.println(number + " is a Prime Number");		
+	}
+	
+	//Calculate Sum, Count and Average of Prime Numbers
+	void primeNumber2To100(int startNum, int endNum) {
```
KrishnaKTechnocredits, line 21: Still, method name is not updated.

AavrutiDJun20: Sorry, Krishna. I missed that one :(
I will keep it in mind from next time

PR #509, opened 2020-06-27
Palak_Assignment19_(40min+10+5+5+10)Min

PalakSJun20:
image
image
image

### DIFF
--- a/src/palakSJun20/DifferIndexFromArray.java
+++ b/src/palakSJun20/DifferIndexFromArray.java
@@ -56,14 +56,19 @@ public class DifferIndexFromArray {
 		DifferIndexFromArray differIndexFromArray = new DifferIndexFromArray();
 		int[] array1 = {10,2,9,14,3};
 		int[] array2 = {10,2,18,14,3};
+		System.out.println("Find first differ index from given arrays Output : ");
 		differIndexFromArray.findFirstDifferIndex(array1 , array2);
+		System.out.println("---------------------------------------------");
 		
 		int[] array3 ={10,2,9,14,3};
 		int[] array4 = {10,2,18,14,13};
+		System.out.println("Find all differ index from given arrays Output : " );
 		differIndexFromArray.findAllDifferIndex(array3, array4); 
+		System.out.println("----------------------------------------------");
 		
 		int[] array5 = {10,2,9,14,3}; 
 		int[] array6 = {10,2,18,14,3};
+		System.out.println("Find max value at differed index Output : ");
 		differIndexFromArray.findMaxValueAtDifferIndex(array5, array6);
 	}
 
--- a/src/palakSJun20/DifferIndexFromArray.java
+++ b/src/palakSJun20/DifferIndexFromArray.java
@@ -1,0 +1,70 @@
+/*Program 2 : Find first differ index from given arrays. 
+                   i/p :- arr1: {10,2,9,14,3}
+                          arr2: {10,2,18,14,3}
+                   o/p:- Values are not matching at index -> 2 
+                   
+Program 3: Find all differ index from given arrays. 
+                 i/p:- arr1: {10,2,9,14,3}
+                        arr2: {10,2,18,14,13}
+                o/p:- Values are not matching at index -> 2
+                         Values are not matching at index -> 4                  
+Program 4: Find max value at differed index.
+              i/p:- arr1: {10,2,9,14,3}
+                     arr2: {10,2,18,14,3}
+              o/p :- Values are not matching at index -> 2, From (9,18) max value is 18
+                                        */
+
+package palakSJun20;
+
+public class DifferIndexFromArray {
+	
+	//Find first differ index from given arrays. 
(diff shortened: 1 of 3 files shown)


src/palakSJun20/DifferIndexFromArray.java
```diff
@@ -0,0 +1,75 @@
...
+	void findFirstDifferIndex(int[] array1 , int[] array2){
+		
+		if (array1.length == array2.length) {
+			for (int i = 0; i < array1.length; i++) {
+					if(array1[i] != array2[i]) {
+						System.out.println("Values are not matching at index " + i);
+						
+				}
+			}
+		}
```
KrishnaKTechnocredits: Where is the else block ???

src/palakSJun20/DifferIndexFromArray.java
```diff
@@ -0,0 +1,75 @@
...
+	void findMaxValueAtDifferIndex(int[] array5, int[] array6){
+		if (array5.length == array6.length) {
+			for (int i = 0; i < array5.length; i++) {
+					if(array5[i] != array6[i]) {
+						System.out.println("Values are not matching at index " + i);
+						if(array5[i] > array6[i])
+							System.out.println("From "+ "(" + array5[i] +","+ array6[i]  + ")" + " Max Value is" + array5[i]);
+						else 
+							System.out.println("From "+ "(" + array5[i] +","+ array6[i]  + ")" + " Max Value is " + array6[i]);
+				}
```
KrishnaKTechnocredits, line 51: Where is the else block ???

src/palakSJun20/DifferIndexFromArray.java
```diff
@@ -0,0 +1,75 @@
...
+	
+	//Find max value at differed index.
+	void findMaxValueAtDifferIndex(int[] array5, int[] array6){
+		if (array5.length == array6.length) {
+			for (int i = 0; i < array5.length; i++) {
+					if(array5[i] != array6[i]) {
+						System.out.println("Values are not matching at index " + i);
+						if(array5[i] > array6[i])
+							System.out.println("From "+ "(" + array5[i] +","+ array6[i]  + ")" + " Max Value is" + array5[i]);
+						else
```
KrishnaKTechnocredits: The printing statement is not as expected.
System.out.println("Values are not matching at index -> " + index + ", From (" + arr1[index] + ","
							+ arr2[index] + ") max value is " + max);
You should write this line after if block.

if (arr1.length == arr2.length) {
			for (int index = 0; index < arr1.length; index++) {
				int max = arr2[index];
				if (arr1[index] != arr2[index]) {
					if (arr1[index] > arr2[index])
						max = arr1[index];
					System.out.println("Values are not matching at index -> " + index + ", From (" + arr1[index] + ","
							+ arr2[index] + ") max value is " + max);
				}
			}
		} else
			System.out.println("Both array's length is different.");

src/palakSJun20/PrimeNumberVariation.java
```diff
@@ -0,0 +1,84 @@
...
+			}
+			if (flag == true) {
+				System.out.println("Number "+number+ " is not prime");
+			}
+			else 
+				System.out.println("Number "+ number+ " is Prime");
+		}
+		
+		//Print Total count of prime number b/w 2-100
+		int printTotalCountOfPrimeNumber(){
```
KrishnaKTechnocredits: Why this separate method is required ? You should directly write count and some logic in findGivenNumberPrimeOrNot() method.  Take two instance variable.

src/palakSJun20/PrimeNumberVariation.java
```diff
@@ -0,0 +1,84 @@
...
+				}
+				if(flag == false){
+					count++;
+				}
+			}
+			return count;
+		}
+		
+		//Find sum of all prime numbers between 2->100
+		int findSumOfPrimeNumbers(){
```
KrishnaKTechnocredits: Not required.

KrishnaKTechnocredits (changes requested):
..

src/palakSJun20/DifferIndexFromArray.java
PalakSJun20: Updated with Else block

src/palakSJun20/DifferIndexFromArray.java
PalakSJun20: updated with else block

src/palakSJun20/DifferIndexFromArray.java
PalakSJun20: as per suggested changes has been updated.

src/palakSJun20/PrimeNumberVariation.java
PalakSJun20: written sum and average logic in same method where we are counting prime number from 2-100

src/palakSJun20/PrimeNumberVariation.java
PalakSJun20: Method Removed

PR #511, opened 2020-06-27
Abhishek_Assignment_19_(40+10+10+10+15)Minutes

AbhishekSJune2020:
Output1:
Count of Prime Number between range :25
Sum of Prime Number between range :1060
Average of Prime Number between range 2 - 100 is: 42.0

Output:2
Values are not matching at index -> 2
 
Output: 3
Values are not matching at index -> 2
Values are not matching at index -> 4		

Output-4
Values are not matching at index -> 2, From (9,18) max value is18

Output:5
Values are not matching at index -> 2, [arr1 value is Krishna, arr2 value is Technocredits]
Values are not matching at index -> 3, [arr1 value is Technocredits, arr2 value is Krishna]

### DIFF
--- a/src/abhishekSJun20/FirstDifferIndexOfArray.java
+++ b/src/abhishekSJun20/FirstDifferIndexOfArray.java
@@ -8,12 +8,13 @@ package abhishekSJun20;
 
 public class FirstDifferIndexOfArray {
 
-	void differIndexValueInIndex(int[] arr1, int[] arr2) {
+	int differIndexValueInIndex(int[] arr1, int[] arr2) {
 
 		for (int index = 0; index < arr1.length; index++) {
 			if (arr1[index] != arr2[index])
 				System.out.println("Values are not matching at index -> " + index);
 		}
+		return 0;
 	}
 
 	public static void main(String[] args) {
--- a/src/abhishekSJun20/FirstDifferIndexOfArray.java
+++ b/src/abhishekSJun20/FirstDifferIndexOfArray.java
@@ -1,0 +1,25 @@
+package abhishekSJun20;
+/*
+ * Program 2: Find first differ index from given arrays. 
+                   i/p :- arr1: {10,2,9,14,3}
+                            arr2: {10,2,18,14,3}
+                   o/p:- Values are not matching at index -> 
+ */
+
+public class FirstDifferIndexOfArray {
+
+	void differIndexValueInIndex(int[] arr1, int[] arr2) {
+
+		for (int index = 0; index < arr1.length; index++) {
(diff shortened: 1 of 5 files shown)


src/abhishekSJun20/FirstDifferIndexOfArray.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+public class FirstDifferIndexOfArray {
+
+	void differIndexValueInIndex(int[] arr1, int[] arr2) {
+
+		for (int index = 0; index < arr1.length; index++) {
+			if (arr1[index] != arr2[index])
+				System.out.println("Values are not matching at index -> " + index);
+		}
```
KrishnaKTechnocredits, line 16: Here you are checking for the first index, why return statement is missing ??

src/abhishekSJun20/PrimeNumber.java
```diff
@@ -0,0 +1,66 @@
...
+		// Method displays and return sum of all the prime number in the range
+		int sum = 0;
+		for (int index = startRange; index <= endRange; index++) {
+			sum += primeNumber(index);
+		}
+		System.out.println("Sum of Prime Number between range :" + sum);
+		return sum;
+	}
+
+	double AverageAllPrimeNumInRange(int sumOfPrimeNum, int countOfPrimeNum) {
```
KrishnaKTechnocredits, line 50: No need to write extra methods, Just create two instance variable sum and count and perform that operation in isPrime method and use those variable in method like:
}

KrishnaKTechnocredits (changes requested):
.

src/abhishekSJun20/PrimeNumber.java
```diff
@@ -0,0 +1,66 @@
...
+
+	int primeNumber(int num) {
+		// Method to find if Number is Prime Number or not. It return num - If prime and
+		// returns 0 - If not Prime
+		for (int index = 2; index <= Math.sqrt(num); index++) {
+			if (num % index == 0) {
+				return 0;
+			}
+		}
+		return num;
```
KrishnaKTechnocredits, line 22: Where is the printing statement, number is prime or not?

src/abhishekSJun20/FirstDifferIndexOfArray.java
AbhishekSJune2020: done changes as per comment.

src/abhishekSJun20/PrimeNumber.java
AbhishekSJune2020: done changes as per comment

PR #514, opened 2020-06-27
Nikhil_Assignment_19_Done_(20+4+4+8+4=)80Min

NikhilMJun20:
->Completed and pushed Assignment # 19.
->Please review and comment.

->PFB screenshots of the outputs of the programs:-
**Adding updated screenshot for PrimeNumber code...**

Annotation 2020-06-28 192734

**1**
Output_PrimeNumbers

**2**
Output_FirstDiffIndex

**3**
Annotation 2020-06-27 224748

**4**
Ouput_MaxValueAtDiffIndex

**5**
Annotation 2020-06-27 224530


src/nikhilMJun20/AllDiffIndex.java
```diff
@@ -0,0 +1,25 @@
...
+arr2: {10,2,18,14,13}
+o/p:- Values are not matching at index -> 2
+Values are not matching at index -> 4
+*/
+
+package nikhilMJun20;
+
+public class AllDiffIndex {
+	void displayAllDiffIndex(int[] arr1, int[] arr2) {
+		for(int index=0; index<arr1.length; index++) {
```
KrishnaKTechnocredits: Nikhil, to write directly for loop is wrong. Because first, you should check that Both arrays length is equal or not. 
If length is not equal then no need to proceed further. 

Please add this change to all your methods.

KrishnaKTechnocredits: When do you write like this " index<arr1.length; "  interview directly ask you that on which basis you decided that length should be as arr1.length why not arr12.length ? 
So initially if you write if (arr1.length == arr2.length) this then you can iterate loop till any arrays length.

src/nikhilMJun20/MaxValueAtDiffIndex.java
```diff
@@ -0,0 +1,28 @@
...
+i/p:- arr1: {10,2,9,14,3}
+arr2: {10,2,18,14,3}
+o/p :- Values are not matching at index -> 2, From (9,18) max value is 18
+*/
+
+package nikhilMJun20;
+
+public class MaxValueAtDiffIndex {
+	void findMaxAtDiffIndex(int[] arr1, int[] arr2) {
+		int max = 0;
```
KrishnaKTechnocredits: No need to initialize here. You can directly initialize at line number 13.

src/nikhilMJun20/PrimeNumbers.java
```diff
@@ -0,0 +1,51 @@
...
+				break;
+			}
+		}
+		if(flag == true)
+			System.out.print(num+" is a Prime Number.");
+		else
+			System.out.println(num+" is NOT a Prime Number.");
+	}
+	
+	void findPimNumbers() {
```
KrishnaKTechnocredits: Here you just created one method 
			isPrime(index);
		} 
And performed all operations like sum, count in isPrime() method, and create two instance variables for the same.

KrishnaKTechnocredits (changes requested):
.

src/nikhilMJun20/AllDiffIndex.java
NikhilMJun20: Hi @KrishnaKTechnocredits,
Made this change to all the array programs.
Thanks!

src/nikhilMJun20/MaxValueAtDiffIndex.java
NikhilMJun20: Hi @KrishnaKTechnocredits,
Made the suggested change.

src/nikhilMJun20/PrimeNumbers.java
NikhilMJun20: Hi @KrishnaKTechnocredits,
Made the necessary changes to the code.
Thanks!

KrishnaKTechnocredits (approved):
Nikhil perfect code. Good job.

NikhilMJun20:
> Nikhil perfect code. Good job.

Thanks, @KrishnaKTechnocredits !

PR #516, opened 2020-06-27
Aastha_Assignment_19_Done_70mins

AasthaJJun20:
->Pushed 2 class file
1) for Program 1
2) for Program 2 to 5
Struggled in printing the average value in decimals. :(
PFB output:
PrimeNumbers-output
TestComparisionOfArrays-output

### DIFF
--- a/src/aasthaJJun20/PrimeNumbers.java
+++ b/src/aasthaJJun20/PrimeNumbers.java
@@ -1,51 +1,38 @@
 package aasthaJJun20;
 
 public class PrimeNumbers {
-	//a) Given number is prime or not. 
-		boolean isPrimeNumber(int num){
-			for(int index = 2; index <= Math.sqrt(num); index++){
-				if(num%index == 0){
-					return false;
-				}
+	boolean isPrimeNumber(int num){
+		for(int index = 2; index <= Math.sqrt(num); index++){
+			if(num%index == 0){
+				return false;
 			}
-			return true;
 		}
-		
-		// b) Print total count of prime numbers between 2->100.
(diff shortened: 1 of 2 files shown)


src/aasthaJJun20/PrimeNumbers.java
```diff
@@ -0,0 +1,51 @@
...
+			for(int index = start; index <= end; index++){
+				if(isPrimeNumber(index))
+					sum += index;
+			}
+			return sum;
+		}
+		
+		//d) Find avg of prime numbers between 2->100
+		double findAvgOfPrimesInRange(int start, int end){
+			return Double.valueOf(findSumOfPrimesInRange(start,end))/Double.valueOf(findTotalPrimeInRange(start,end));
```
KrishnaKTechnocredits: You just need to take two instance variable one for Sum and count.
And implement both logic in isPrimeNumber() method then no need to write an extra two methods. 
ANd printing statement itself you can calculate sum/count = avg.

KrishnaKTechnocredits (changes requested):
.

src/aasthaJJun20/PrimeNumbers.java
AasthaJJun20: Hi Krishna,
Sorry I misinterpreted the requirement, I thought we have to write a method for each mentioned point.
As per your comment, I have removed 3 methods for - sum, count, average and also removed the if-else statements from main method so that the whole logic stays in display methods. Please review further.
Thanks.
PrimeNumbers-output

KrishnaKTechnocredits (approved):
Good job. Now perfect code.

PR #517, opened 2020-06-27
Vaishnavi_Assignment_19_80MIN_Done

VaishnaviVJun20:
image
image
image
image
image


src/vaishnaviVJun20_arrayExamples/FirstDifferIndex.java
```diff
@@ -0,0 +1,19 @@
+package vaishnaviVJun20_arrayExamples;
+
+public class FirstDifferIndex {
+	
+	void findFirstDifferIndex(int[] array1, int[] array2) {
+	   for(int index=0;index<array1.length;index++) {
+		   if(array1[index]!= array2[index])
+			   System.out.println("Values are not matching at index->" +index);
```
KrishnaKTechnocredits: Where is the return statement ?

src/vaishnaviVJun20_arrayExamples/FirstDifferIndex.java
```diff
@@ -0,0 +1,19 @@
...
+	   }	
+	}
+
+	public static void main(String[] args) {
+		FirstDifferIndex firstDifferIndex = new FirstDifferIndex();
+		int[] array1 = {10,2,9,14,3};
+		int[] array2 = {10,2,18,14,3};
```
KrishnaKTechnocredits: If here arrays value changed then your code is not working without return statement.

src/vaishnaviVJun20_arrayExamples/Prime.java
```diff
@@ -0,0 +1,66 @@
...
+			 flag= false;
+			 System.out.println(input+" is not a prime number");
+			 break;
+		   }
+		}
+		if(flag == true)
+			System.out.println(input +" is a Prime number");
+	}
+	
+	void findPrimeNumbers() {
```
KrishnaKTechnocredits: This method should be parameterized. If tomorrow I want to use this method for range 100 to 1000, this won't work.

src/vaishnaviVJun20_arrayExamples/Prime.java
```diff
@@ -0,0 +1,66 @@
...
+		    }
+		    if(flag == true)
+		    	sum +=index;
+		    			    
+		 } 
+		 System.out.println("count of prime numbers between 2 to 100 is: " +sum);
+		
+	}
+	
+	void findAvgOfAllPrimeNumbers() {
```
KrishnaKTechnocredits: No need to write extra methods, Just create two instance variable sum and count and perform that operation in isPrime method and use those variable in method like:
		}

KrishnaKTechnocredits (changes requested):
.

src/vaishnaviVJun20_arrayExamples/FirstDifferIndex.java
VaishnaviVJun20: I have added the return statement. Got the logic. Thanks for the suggestion.

src/vaishnaviVJun20_arrayExamples/FirstDifferIndex.java
VaishnaviVJun20: Done

src/vaishnaviVJun20_arrayExamples/Prime.java
VaishnaviVJun20: Done suggested changes.

src/vaishnaviVJun20_arrayExamples/Prime.java
VaishnaviVJun20: done suggested changes. Thanks for the logic suggestion.

KrishnaKTechnocredits (approved):
Good job dear.

PR #518, opened 2020-06-27
Raj_Assignment_19_90min_Done

RajAJun20:
Assignment_19_alldifferString
Assignment_19_AllIndex
Assignment_19_FirstDifferIndex
Assignment_19_Max
Assignment_19_PrimeNumber

### DIFF
--- a/src/rajAJun20/AllDifferIndexString.java
+++ b/src/rajAJun20/AllDifferIndexString.java
@@ -1,0 +1,29 @@
+package rajAJun20;
+
+public class AllDifferIndexString {
+	
+	void allDifferIndexOfStringArray(String[] arr1 , String[] arr2){
+		
+		if(arr1.length == arr2.length) {
+			for(int index = 0 ; index < arr1.length ; index++ ) {
+				if(arr1[index] != arr2[index]) {
+					System.out.println("String Values are not matching at index -> " + index + " , [arr1 value is " + arr1[index] + ", arr2 value is " + arr2[index] + "]" );
+				}
+			}
+		}
+		else {
+			System.out.println("Both array lengths are not equal or identical");
+		}
+	}
+	
+	public static void main(String[] args) {
+		AllDifferIndexString allDifferStringIndex = new AllDifferIndexString();
+	 
+		String[] string1 = {"Maulik","Harsh","Krishna","Technocredits"};
+		String[] string2 = {"Maulik","Harsh","Technocredits","Krishna"};
+		
+		allDifferStringIndex.allDifferIndexOfStringArray(string1, string2);
+		
+	}
+	
+}
--- a/src/rajAJun20/AllDifferIndexString.java
+++ b/src/rajAJun20/AllDifferIndexString.java
@@ -4,9 +4,10 @@ public class AllDifferIndexString {
 	
 	void allDifferIndexOfStringArray(String[] arr1 , String[] arr2){
 		
-		if(arr1.length == arr2.length) {
+		if(arr1.length == arr2.length) {       //if (!s1[i].equals(s2[i]))
+
 			for(int index = 0 ; index < arr1.length ; index++ ) {
-				if(arr1[index] != arr2[index]) {
+				if(!arr1[index].equals(arr2[index])) {
 					System.out.println("String Values are not matching at index -> " + index + " , [arr1 value is " + arr1[index] + ", arr2 value is " + arr2[index] + "]" );
 				}
 			}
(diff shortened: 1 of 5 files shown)


src/rajAJun20/AllDifferIndexString.java
```diff
@@ -0,0 +1,29 @@
+package rajAJun20;
+
+public class AllDifferIndexString {
+	
+	void allDifferIndexOfStringArray(String[] arr1 , String[] arr2){
+		
+		if(arr1.length == arr2.length) {
+			for(int index = 0 ; index < arr1.length ; index++ ) {
+				if(arr1[index] != arr2[index]) {
```
KrishnaKTechnocredits: For string comparison you should always use .equals() method.

src/rajAJun20/FirstDifferIndex.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class FirstDifferIndex {
+	
+	void differFirstIndexOfArrays(int[] arr1 , int[] arr2 ) {
+		
+		if(arr1.length == arr2.length) {
+				for(int index = 0 ; index < arr1.length ; index++ ) {
+					if(arr1[index] != arr2[index]) {
+						System.out.println("Numbers of both the arrays are not matching at index : " + index);
+					}
```
KrishnaKTechnocredits, line 12: Where is the return statement ??? differFirstIndexOfArrays() we are saying on first index what if other index value is also different.

src/rajAJun20/MaximumDifferIndexValue.java
```diff
@@ -0,0 +1,27 @@
+package rajAJun20;
+
+public class MaximumDifferIndexValue {
+	
+	void maxValueOfDifferIndex(int[] arr1 , int[] arr2) {
+		if(arr1.length == arr2.length) {
+			for(int index = 0 ; index < arr1.length ; index++) {
+				if(arr1[index] != arr2[index]) {
+					System.out.println("Numbers are not matching at index : " + index);
+				if(arr1[index] > arr2[index])
```
KrishnaKTechnocredits: Again this is wrong. Here you should write second if inside first if.

src/rajAJun20/PrimeNumberOperations.java
```diff
@@ -0,0 +1,66 @@
...
+			}
+		}
+		if (flag == true) {
+			System.out.println("Number "+number+ " is not Prime");
+		}
+		else 
+			System.out.println("Number "+ number+ " is Prime");
+	}
+	
+	int totalCountOfPrimeNumbers(){
```
KrishnaKTechnocredits, line 24: Where is the range ? Method should be parameterized.

src/rajAJun20/PrimeNumberOperations.java
```diff
@@ -0,0 +1,66 @@
...
+				}
+			}
+			if (flag == false) {
+				sum = sum + i;
+			}
+		}
+		return sum;
+	}
+
+	void avgOfPrimeNumbers(){
```
KrishnaKTechnocredits, line 54: No need to write extra methods, Just create two instance variable sum and count and perform that operation in isPrime method and use those variable in method like:
}

KrishnaKTechnocredits (changes requested):
.

src/rajAJun20/PrimeNumberOperations.java
```diff
@@ -0,0 +1,66 @@
...
+		}
+		return count;
+	}
+
+	int sumOfPrimeNumbers(){
+		int sum = 0;
+		for (int i = 2; i < 100; i++) {
+			boolean flag = false;
+			for (int j = 2; j < i; j++) {
+				if ( i%j == 0) {
```
KrishnaKTechnocredits: Here you should use numberIsPrimeOrNot() method.

src/rajAJun20/PrimeNumberOperations.java
```diff
@@ -0,0 +1,66 @@
...
+		else 
+			System.out.println("Number "+ number+ " is Prime");
+	}
+	
+	int totalCountOfPrimeNumbers(){
+		int count = 0;
+		for (int i = 2; i < 100; i++) {
+			boolean flag = false;
+			for (int j = 2; j < i ; j++) {
+				if(i % j == 0 ) {
```
KrishnaKTechnocredits: Here you should use numberIsPrimeOrNot() method.

src/rajAJun20/AllDifferIndexString.java
RajAJun20: Done

src/rajAJun20/FirstDifferIndex.java
RajAJun20: Used break now

src/rajAJun20/MaximumDifferIndexValue.java
RajAJun20: Done

src/rajAJun20/PrimeNumberOperations.java
RajAJun20: Changed method to print number is prime or not and called primeNumber method to show count and sum of prime numbers

PR #520, opened 2020-06-27
Swapna_Assignment_19_Done_60+10+5+5+10

SwapnaWJun20:
image
image
image
image
image

### DIFF
--- a/src/swapnaWJun20/PrimeOperations.java
+++ b/src/swapnaWJun20/PrimeOperations.java
@@ -8,18 +8,13 @@ package swapnaWJun20;
 
 public class PrimeOperations {
 
-	void totalCount(int a, int b) {
+	void findPrimeNumberInGivenRange(int first, int last) {
+
 		int sum = 0;
 		int count = 0;
-		for (int start = a; start < b; start++) {
-			boolean flag = false;
-			for (int index = 2; index <= Math.sqrt(start); index++) {
-				if (start % index == 0) {
-					flag = true;
-					break;
-				}
-			}
-			if (flag == false) {
+		for (int start = first; start <= last; start++) {
+			boolean flag = isPrime(start);
+			if (flag == true) {
 				count++;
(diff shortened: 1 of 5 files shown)


src/swapnaWJun20/PrimeOperations.java
```diff
@@ -0,0 +1,53 @@
...
+                  a) Given number is prime or not. 
+                  b) Print total count of prime numbers between 2->100.
+                  c) Find sum of all prime numbers between 2->100
+                  d) Find avg of prime numbers between 2->100*/
+
+package swapnaWJun20;
+
+public class PrimeOperations {
+
+	void totalCount(int a, int b) {
```
KrishnaKTechnocredits: Method name is not expected like this... findPrimeNumberInGivenRange(first,last)

src/swapnaWJun20/PrimeOperations.java
```diff
@@ -0,0 +1,53 @@
...
+		int sum = 0;
+		int count = 0;
+		for (int start = a; start < b; start++) {
+			boolean flag = false;
+			for (int index = 2; index <= Math.sqrt(start); index++) {
+				if (start % index == 0) {
```
KrishnaKTechnocredits: Why are you writing whole logic here, call isPrime() method.

KrishnaKTechnocredits (changes requested):
.

src/swapnaWJun20/PrimeOperations.java
SwapnaWJun20: @KrishnaKTechnocredits Changes done as per the comments.

src/swapnaWJun20/PrimeOperations.java
SwapnaWJun20: @KrishnaKTechnocredits  Changes done as per the comments.

PR #523, opened 2020-06-27
Archana_Assignment_19_70min

ArchanaNJun20:
PrimeNumberOpration_40min
image
FirstDifferIndex_8min
image
AllDifferIndex_8min
image
MaxDiffer_8min
image
AlldDifferString_8min
image

### DIFF
--- a/src/archanaNJun20/PrimeNumberOpration.java
+++ b/src/archanaNJun20/PrimeNumberOpration.java
@@ -1,0 +1,46 @@
+package archanaNJun20;
+
+import java.util.Scanner;
+
+public class PrimeNumberOpration {
+	void primeOperation(){
+		int count=0;
+		int sum=0;
+		System.out.println("Prime Numbers between 2->100 :");
+		for(int index=2;index<=100;index++)	{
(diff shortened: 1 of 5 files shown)


KrishnaKTechnocredits (changes requested):
Seriously Archana, All method name is the same, So why you have created so many methods just call only one methods. Nobody going to check your code if you have written like this.

src/archanaNJun20/PrimeNumberOpration.java
```diff
@@ -0,0 +1,46 @@
+package archanaNJun20;
+
+import java.util.Scanner;
+
+public class PrimeNumberOpration {
+	void primeOperation(){
```
KrishnaKTechnocredits: make this method parameterized.

ArchanaNJun20: The code is updated based on the given comments.
method names changed, from next time onwards  I will take care while writing.

KrishnaKTechnocredits (approved):
Now cod looks perfect, Thank you for considering all the changes. Good job.

PR #527, opened 2020-06-27
ADT_Assignment_19_Done_50Mins

AditiGJUN20:
**Assignment 19:**
1. **PrimeNumberOps.java**
Time: 25 mins
Screen Shot 2020-06-27 at 2 48 49 PM

2. **FirstDifferIndexInArray.java**
Time: 5 Mins
Screen Shot 2020-06-27 at 4 09 08 PM

3. **AllDifferIndexInArray.java**
Time: 5 Mins
Screen Shot 2020-06-27 at 4 10 10 PM

4. **FindMaxAtDifferIndex.java**
Time: 8Mins
Screen Shot 2020-06-27 at 4 12 50 PM

5. **AllDifferIndexInStringArray.java**
Time: 8Mins
Screen Shot 2020-06-27 at 4 13 05 PM

### DIFF
--- a/src/aditiGJUN20/AllDifferIndexInArray.java
+++ b/src/aditiGJUN20/AllDifferIndexInArray.java
@@ -11,7 +11,7 @@ package aditiGJUN20;
 public class AllDifferIndexInArray {
 	
 	// Method to find array index which has different element value 
-		void getAllDifferIndex(int[] arr1, int[] arr2) {
+		void findAllDifferIndex(int[] arr1, int[] arr2) {
 			if(arr1.length==arr2.length) {
 				for(int index=0;index<arr1.length;index++) {
 					if(arr1[index]!=arr2[index])
@@ -28,6 +28,6 @@ public class AllDifferIndexInArray {
 			System.out.println("All differ Index: ");
 			System.out.println("arr1={10,2,9,14,3}\n" + 
 					"arr2={10,2,18,14,13}");
-			allDifferIndexInArray.getAllDifferIndex(arr1, arr2);
+			allDifferIndexInArray.findAllDifferIndex(arr1, arr2);
 		}
 }
--- a/src/aditiGJUN20/AllDifferIndexInArray.java
+++ b/src/aditiGJUN20/AllDifferIndexInArray.java
@@ -1,0 +1,33 @@
+/*
+Program 3: Find all differ index from given arrays. 
+	i/p:- arr1: {10,2,9,14,3}
+    arr2: {10,2,18,14,13}
+    o/p:- Values are not matching at index -> 2
+    Values are not matching at index -> 4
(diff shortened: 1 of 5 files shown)


src/aditiGJUN20/AllDifferIndexInArray.java
```diff
@@ -0,0 +1,33 @@
...
+    o/p:- Values are not matching at index -> 2
+    Values are not matching at index -> 4
+*/        
+
+package aditiGJUN20;
+
+public class AllDifferIndexInArray {
+	
+	// Method to find array index which has different element value 
+		void getAllDifferIndex(int[] arr1, int[] arr2) {
```
KrishnaKTechnocredits: Aditi when we start method name as get it means that we are going to return something from that method. 
So please don't use that. Instead of it, you can use findAllDifferIndex() as method name.

src/aditiGJUN20/FindMaxAtDifferIndex.java
```diff
@@ -0,0 +1,35 @@
...
+	i/p:- arr1: {10,2,9,14,3}
+	arr2: {10,2,18,14,3}
+	o/p :- Values are not matching at index -> 2, From (9,18) max value is 18 
+ * */
+
+package aditiGJUN20;
+
+public class FindMaxAtDifferIndex {
+	void getMaxAtDifferIndex(int[] arr1, int[] arr2) {
+		int max=0;
```
KrishnaKTechnocredits: No need to initialize here. You should directly initialize at line number 15.

src/aditiGJUN20/PrimeNumberOps.java
```diff
@@ -0,0 +1,62 @@
...
+		if(flag==true)
+			System.out.println(num+" is Prime Number.");
+		else 
+			System.out.println(num+" is not Prime Number.");
+	}
+	
+	// Method to find prime numbers in given range.
+	void arithmaticOpsOnPrimeNum() {
+		boolean flag=false;
+		System.out.println("\nPrime numbers within 2-100 range: ");
```
KrishnaKTechnocredits, line 38: No need to write extra methods, Just create two instance variable sum and count and perform that operation in isPrime method and use those variable in method like:
isPrime(index);
}

src/aditiGJUN20/AllDifferIndexInArray.java
AditiGJUN20: Method name changed as suggested.

src/aditiGJUN20/FindMaxAtDifferIndex.java
AditiGJUN20: Made changes. 
Initialized the variable where I'm assigning the value.

PR #529, opened 2020-06-27
Anup_Assignment_19_Done_60mins

AnupSJun20:
Anup_Assignment_19_Done_60mins_Please review.
Program1
Program2
Program3
Program4
Program5

### DIFF
--- a/src/anupSJun20/AllDifferIndex.java
+++ b/src/anupSJun20/AllDifferIndex.java
@@ -1,0 +1,33 @@
+package anupSJun20;
+/* Program 5: Find all differ index from given arrays.
+               i/p:- arr1: {"Maulik", "Harsh", "Krishna", "Technocredits"}
+                       arr2: {"Maulik","Harsh","Technocredits","Krishna"}		
+               o/p:- Values are not matching at index -> 2, [arr1 value is Krishna, arr2 value is Technocredits] 
+                        Values are not matching at index -> 3, [arr1 value is Technocredits, arr2 value is Krishna] */
+
+public class AllDifferIndex {
+	static void findAllDifferIndex(String[] arr1,String[] arr2) {
+		boolean flag = true;
+		if(arr1.length == arr2.length) {
+			for(int index = 0;index<arr1.length;index++) {
+				if(!arr1[index].equalsIgnoreCase(arr2[index])) {
+						flag = false;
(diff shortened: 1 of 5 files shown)


src/anupSJun20/AllDifferIndex.java
```diff
@@ -0,0 +1,33 @@
...
+	static void findAllDifferIndex(String[] arr1,String[] arr2) {
+		boolean flag = true;
+		if(arr1.length == arr2.length) {
+			for(int index = 0;index<arr1.length;index++) {
+				if(!arr1[index].equalsIgnoreCase(arr2[index])) {
+						flag = false;
+						System.out.println("Values are not matching at index -> " +index+ " arr1 value is "+ arr1[index] +","+ " arr2 value is "+ arr2[index]);
+					}
+				}
+			if (flag == true) {
```
KrishnaKTechnocredits, line 18: No need to use this variable here, because our motto is to check findAllDifferIndex. Please apply this change in every method.

src/anupSJun20/FindAllDifferIndex.java
```diff
@@ -0,0 +1,31 @@
...
+		boolean flag = true;
+		if(arr1.length == arr2.length) {
+		for(int index = 0;index<arr1.length;index++) {
+			if(arr1[index]!=arr2[index]) {
+				flag = false;
+				System.out.println("Values are not matching at "+ index);
+			}
+		}
+		if(flag ==true) {
+			System.out.println("");
```
KrishnaKTechnocredits, line 19: This does not looks good.

src/anupSJun20/PrimeNumber.java
```diff
@@ -0,0 +1,53 @@
...
+	
+	static int sumOfPrimeNumbers(int startNum, int endNum) {
+		int sum = 0;
+		for(int index = startNum;index<=endNum;index++) {
+			if (findPrimeNumber(index) != 0)
+				sum = sum + findPrimeNumber(index);
+		}
+		return sum;
+	}
+	static void findaverageofPrimeNumbers(int startNum, int endNum) {
```
KrishnaKTechnocredits, line 33: No need to write extra methods, Just create two instance variable sum and count and perform that operation in isPrime method and use those variable in method like:
  findPrimeNumber(index);
}
System.out.println("Average of all prime number is: "+sum/count)

Also you can avoid return statement if you want. Set logic as you want.

KrishnaKTechnocredits (approved):
Good job, Please ignore ex comments.

PR #531, opened 2020-06-28
Shruti Assignment 19 done 40mints

ShrutiDJun20:
outputs: Find all differ index from given arrays
image

o/p - Find max value at differed index.
image
 o/pFind all differ index from given arrays.
image

Find first differ index from given arrays
image

### DIFF
--- a/src/shrutiDJun20/CompareIntArray.java
+++ b/src/shrutiDJun20/CompareIntArray.java
@@ -1,0 +1,32 @@
+/*
+ Find first differ index from given arrays. 
+arr1: {10,2,9,14,3}
+arr2: {10,2,18,14,3}
+o/p:- Values are not matching at index -> 2 
+ */
+package shrutiDJun20;
+
+public class CompareIntArray {
+
+	void compareArray(int[] input1, int[] input2){
+		//compare length first
+		int diffIndex=0;
+		if(input1.length==input2.length){
+			for(int i=0;i<input1.length;i++){
+				if(input1[i]!= input2[i]){
+					diffIndex=i;
+					System.out.println("first difference found in given arrays is at index "+ diffIndex);
+					return;
+				}
+			}
+		}
+	}
+		
+	
+	public static void main(String[] args) {
+		CompareIntArray compareIntArray = new CompareIntArray();
+		int input1[] = {10,2,9,14,3};
+		int input2[]= {10,2,18,14,3};
+		compareIntArray.compareArray(input1, input2);
+	}
+}
--- a/src/shrutiDJun20/FindAllDiff.java
+++ b/src/shrutiDJun20/FindAllDiff.java
@@ -1,22 +1,38 @@
 package shrutiDJun20;
 
 public class FindAllDiff {
-	
-	void findDiffInArr(int[] arr1, int[] arr2){
+
+	void findDiffInArr(int[] arr1, int[] arr2) {
 		int temp = 0;
-		if(arr1.length==arr2.length){
-			for(int index=0;index<arr1.length;index++){
-				if(arr1[index]!=arr2[index]){
-					temp =index;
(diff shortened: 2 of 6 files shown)


src/shrutiDJun20/CompareIntArray.java
```diff
@@ -0,0 +1,32 @@
...
+		//compare length first
+		int diffIndex=0;
+		if(input1.length==input2.length){
+			for(int i=0;i<input1.length;i++){
+				if(input1[i]!= input2[i]){
+					diffIndex=i;
+					System.out.println("first difference found in given arrays is at index "+ diffIndex);
+					return;
+				}
+			}
```
KrishnaKTechnocredits, line 21: Where is the else statement ?

src/shrutiDJun20/CompareIntArray.java
```diff
@@ -0,0 +1,32 @@
...
+ Find first differ index from given arrays. 
+arr1: {10,2,9,14,3}
+arr2: {10,2,18,14,3}
+o/p:- Values are not matching at index -> 2 
+ */
+package shrutiDJun20;
+
+public class CompareIntArray {
+
+	void compareArray(int[] input1, int[] input2){
```
KrishnaKTechnocredits, line 11: Your method name sounds like you are comparing whole array. Your method name should be findDifferIndexOfGivenArray() like this.

src/shrutiDJun20/FindAllDiff.java
```diff
@@ -0,0 +1,22 @@
...
+	void findDiffInArr(int[] arr1, int[] arr2){
+		int temp = 0;
+		if(arr1.length==arr2.length){
+			for(int index=0;index<arr1.length;index++){
+				if(arr1[index]!=arr2[index]){
+					temp =index;
+				System.out.println("Values are not matching at index ->"+temp);
+				}
+			}
+		}
```
KrishnaKTechnocredits, line 26: Here is also else statement missing.

src/shrutiDJun20/CompareIntArray.java
KrishnaKTechnocredits: for example, array length is not equal then ? At least tell use that given two arrays are of different length.

src/shrutiDJun20/StringArrIndexDiff.java
```diff
@@ -0,0 +1,34 @@
...
+	
+	void findDiffStringIndex(String[] input1, String[] input2){
+		String s1 ="";
+		String s2 ="";
+		if(input1.length==input2.length){
+		for(int i = 0 ; i < input1.length ; i++){
+		//	for(int j = 0 ; j < input2.length ; j++){
+				if(!input1[i].equals(input2[i])){
+					s1 = input1[i];
+					s2 = input2[i];
```
KrishnaKTechnocredits: Why you have used this s1 and s2 extra variable. You should directly print  input1[i] in syso statement.

src/shrutiDJun20/CompareIntArray.java
ShrutiDJun20: Changes done as per your comment. Added else statement

src/shrutiDJun20/CompareIntArray.java
ShrutiDJun20: changed method name and added else statement too

src/shrutiDJun20/FindAllDiff.java
ShrutiDJun20: added else statement as per comment

src/shrutiDJun20/StringArrIndexDiff.java
ShrutiDJun20: removed extra variables

PR #533, opened 2020-06-28
Piyush_Assignment_18_1.5Hours

PiyushPJun20:
image

image

image

image

### DIFF
--- a/src/CountSpChar.java
+++ b/src/CountSpChar.java
@@ -2,17 +2,10 @@
 public class CountSpChar {
 	void displaySpCharCount(String str) {
 		String str1 = str;
-
-		str1 = str1.replace('[', ' ');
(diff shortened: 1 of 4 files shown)


src/CountSpChar.java
```diff
@@ -0,0 +1,24 @@
+
+public class CountSpChar {
+	void displaySpCharCount(String str) {
+		String str1 = str;
+
+		str1 = str1.replace('[', ' ');
+		str1 = str1.replace('{', ' ');
```
KrishnaKTechnocredits: replace method can be called in a chain like below code.

str1.replace("[","").replace("{","") so on...

KrishnaKTechnocredits: Code changes based on given your requirement.

PR #534, opened 2020-06-28
Shruti code3 5mints

ShrutiDJun20:
image

### DIFF
--- a/src/shrutiDJun20/SumIs7.java
+++ b/src/shrutiDJun20/SumIs7.java
@@ -1,0 +1,36 @@
+/*Find all pairs on integer array whose sum is equal to given number?
+Hint: You have given an integer array and a number, you need to write a program to find
+all elements in the array whose sum is equal to the given number. Remember, the array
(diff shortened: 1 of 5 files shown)


src/shrutiDJun20/SumIs7.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class SumIs7 {
+	
+	void sumEqualsSeven(int[] arr){
+		//int sum = 0;
+		System.out.println("Integer numbers, whose sum is equal to value:7");
+		for(int index=0;  index<arr.length; index++)
+			
+			for(int index2=0;  index2<arr.length; index2++){
+				if(arr[index]+arr[index2]==7){
```
harshvegada, line 22: This is correct but to safer side always do () for mathematics operation in java

ShrutiDJun20: Ok Harsh.. will keep in mind.. thank you.

PR #535, opened 2020-06-28
Aastha_Coding_Exam_3_5mins

AasthaJJun20:
My laptop is very slow thats why took time to raise PR
IntegerPairSum-output

### DIFF
--- a/src/aasthaJJun20/IntegerPairSum.java
+++ b/src/aasthaJJun20/IntegerPairSum.java
@@ -1,0 +1,19 @@
+package aasthaJJun20;
+
+public class IntegerPairSum {
+	void findPairs(int[] input, int sum){
+		System.out.println("Integer numbers, whose sum is equal to value: "+sum);
+		for(int index=0; index<input.length; index++){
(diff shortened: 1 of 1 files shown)


src/aasthaJJun20/IntegerPairSum.java
```diff
@@ -0,0 +1,19 @@
+package aasthaJJun20;
+
+public class IntegerPairSum {
+	void findPairs(int[] input, int sum){
+		System.out.println("Integer numbers, whose sum is equal to value: "+sum);
+		for(int index=0; index<input.length; index++){
+			for(int innerIndex = 0; innerIndex < input.length; innerIndex++){
+				if(sum == input[index]+input[innerIndex])
```
harshvegada, line 8: This is correct but to safer side always do () for mathematics operation in java

harshvegada (approved):
Excellent work @AasthaJJun20 
Keep it up!

inner for should start from i like below
				if ((a[i] + a[j]) == sum) {
				}
			}
		}

Will discuss in call.
but this also good

AasthaJJun20:
You're right, this will give me more precise output.
will take care of this next time. Thanks

PR #537, opened 2020-06-28
Viresh_Coding_Exam_3_Done_5Mins

VireshJJun20:
Please review and comment.
image

optimized output:
image


harshvegada (approved):
Excellent job @VireshJJun20 

This is correct but to safer side always do () for mathematics operation in java, in your code line number 9

inner for should start from i like below (think on this)
		if ((a[i] + a[j]) == sum) {
		}
	}
}
Current logic is also good!

PR #539, opened 2020-06-28
Barkha_ClassTest3_7mins

BarkhaPJun20:
Find all pairs on integer array whose sum is equal to given number?
Hint: You have given an integer array and a number, you need to write a program to find
all elements in the array whose sum is equal to the given number. Remember, the array
may contain both positive and negative numbers, so your solution should consider that.
Input:2, 4, 3, 5, 7, 8, 9 ,-1
Given sum:7
Integer numbers, whose sum is equal to value:7
(2, 5)
(4, 3)
(5, 2)
(3, 4)
(8, -1)
(-1, 8)
image

### DIFF
--- a/src/barkhaPJun20/PairInArrayHavingDesireSum.java
+++ b/src/barkhaPJun20/PairInArrayHavingDesireSum.java
@@ -1,0 +1,20 @@
+package barkhaPJun20;
+
+public class PairInArrayHavingDesireSum {
+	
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Excellent work @BarkhaPJun20 
Keep it up!

inner for loop should start from i+1 like below (think on this else we discuss in call)
		if ((a[i] + a[j]) == sum) {
		}
	}
}

Current logic is also good!

PR #541, opened 2020-06-28
Palak_ExamCode_3_12Min

PalakSJun20:
image

### DIFF
--- a/src/palakSJun20/ArrayPairSum.java
+++ b/src/palakSJun20/ArrayPairSum.java
@@ -2,10 +2,11 @@ package palakSJun20;
 
 public class ArrayPairSum {
 	void findArrayPairSum(int[] array){
+		System.out.println("Integer numbers, whose sum is equal to value:7");
 		for (int i = 0; i < array.length; i++) {
 			for (int j = 0; j < array.length; j++) {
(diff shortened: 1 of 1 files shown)


PalakSJun20:
image
Sorry for so many commits , firstly i thought i would not able to remove the repetition of Integer then i tried it works :)

harshvegada (approved):
Good job @PalakSJun20 
Keep it up
inner for should start from i like below (think on this else we discuss in cal)

Current is also correct no changes!
No Comments

src/palakSJun20/ArrayPairSum.java
```diff
@@ -0,0 +1,24 @@
+package palakSJun20;
+
+public class ArrayPairSum {
+	void findArrayPairSum(int[] array) {
+		System.out.println("Integer numbers, whose sum is equal to value:7");
+		for (int i = 0; i < array.length; i++) {
+			if (!(array[i] == '*')) {
```
harshvegada, line 7: what is meaning of * here?
could please explain.

PalakSJun20: 1st time when it enters in if condition  replace the the  same element with * so next time loop iterate it will check if * is there it won't go inside.

PR #543, opened 2020-06-28
Jagadeesh_CodeTest3_10min

JagadeeshMJun20:
Please review code.
Output:
image


harshvegada (approved):
Good job @JagadeeshMJun20 
you are near to point but missed index+1 for inner for loop

this also good
keep improving your self

PR #547, opened 2020-06-28
Test3_Done

ApurvaJun20:
image

Output for the program

### DIFF
--- a/src/apurvaJun20/Test3.java
+++ b/src/apurvaJun20/Test3.java
@@ -1,0 +1,22 @@
+package apurvaJun20;
+
+public class Test3 {
+
+	void printPairs(int arr[], int n, int sum) {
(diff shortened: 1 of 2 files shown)


src/apurvaJun20/Test3.java
```diff
@@ -0,0 +1,22 @@
+package apurvaJun20;
+
+public class Test3 {
+
+	void printPairs(int arr[], int n, int sum) {
+		int count = 0;
+		for (int index = 0; index < n; index++)
```
harshvegada, line 7: logic wise good,you got the point in logic but why client will tell you length at calling it is our service method responsibility to check that right?

this is also good.
add this in your improvement part.

harshvegada (approved):
Work on comments next time

PR #548, opened 2020-06-28
Rachana_Coading_Exam_3_10min

RachanaGJun20:
exam3


harshvegada (approved):
Good job @RachanaGJun20 
Keep it up!

		if ((a[i] + a[j]) == sum) {
		}
	}
}

Current logic is also good!
No comments on logic

RachanaGJun20:
Thank you :)

yes got it..... using this logic repetition of pairs will not get appear.

PR #549, opened 2020-06-28
Vaishnavi_Coding_Exam_3_10Min_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20_arrayExamples/PracticeExamTwo.java
+++ b/src/vaishnaviVJun20_arrayExamples/PracticeExamTwo.java
@@ -1,0 +1,23 @@
+package vaishnaviVJun20_arrayExamples;
+
+public class PracticeExamTwo {
+	
+	void findSumEqualToGivenNumber(int[] array1,int num) {
+		System.out.println("Integer numbers, whose sum is equal to value: "+num);
+		 for(int index=0;index<array1.length;index++) {
+			for(int i=1;i<array1.length;i++) {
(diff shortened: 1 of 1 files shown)


src/vaishnaviVJun20_arrayExamples/PracticeExamTwo.java
```diff
@@ -0,0 +1,23 @@
+package vaishnaviVJun20_arrayExamples;
+
+public class PracticeExamTwo {
+	
+	void findSumEqualToGivenNumber(int[] array1,int num) {
+		System.out.println("Integer numbers, whose sum is equal to value: "+num);
+		 for(int index=0;index<array1.length;index++) {
+			for(int i=1;i<array1.length;i++) {
+			   if(array1[index]+array1[i] == num) {
```
harshvegada, line 9: This is correct but to safer side always do () for mathematics operation in java, in your code line number 9

harshvegada (approved):
Good job @VaishnaviVJun20 
Keep it up!

		if ((a[i] + a[j]) == sum) {
		}
	}
}

Current logic is also good!
NO comments on logic everything looks good

src/vaishnaviVJun20_arrayExamples/PracticeExamTwo.java
VaishnaviVJun20: Thank you. I will keep this in mind.

PR #553, opened 2020-06-28
Swapna_Coding_Exam_3_Done_10_min

SwapnaWJun20:
image


harshvegada (approved):
This is correct but to safer side always do () for mathematics operation in java, in your code line number 22

No comments on logic everything looks good!


		if ((a[i] + a[j]) == sum) {
		}
	}
}

PR #554, opened 2020-06-28
Suparna_CodeExam_3_10Min

SuparnaNikam:
Capture


harshvegada (approved):
Good job @SuparnaNJune20 
You are very near to exact logic point this is also correct logic!
Keep it up!
No commments on logic

PR #556, opened 2020-06-28
Deavina_Coding_Exam_3_Done_5_min

DeavinaSJun20:
image


src/deavinaSJun20/PairSumInArray.java
```diff
@@ -0,0 +1,18 @@
...
+
+	public static void main(String[] args) {
+		int[] Arr = { 2, 4, 3, 5, 7, 8, 9, -1 };
+		new PairSumInArray().pairSumInArray(Arr);
+	}
+
+	void pairSumInArray(int[] Arr) {
+		for (int outerindex = 0; outerindex < Arr.length; outerindex++) {
+			for (int innerindex = outerindex + 1; innerindex < Arr.length; innerindex++) {
+				if (Arr[outerindex] + Arr[innerindex] == 7)
```
harshvegada, line 13: This is correct but to safer side always do () for mathematics operation in java, in your code line number 13

logic is correct very efficient!
Keep it up

DeavinaSJun20: Sure Harsh Will keep this in mind for maths operations in future programs

PR #557, opened 2020-06-28
Piyush_Coding_Exam3

PiyushPJun20:
image


harshvegada (approved):
This is correct but to safer side always do () for mathematics operation in java, in your code line number 7

NO comments on logic everything looks good!
Keep it up!

PR #562, opened 2020-06-28
Madhura_Coding_Exam_3_30mins

MadhuraMJun20:
image

### DIFF
--- a/src/madhuraMJun20/scannerClass/Exam3.java
+++ b/src/madhuraMJun20/scannerClass/Exam3.java
@@ -1,0 +1,38 @@
+package madhuraMJun20.scannerClass;
+
+/*Find all pairs on integer array whose sum is equal to given number?
+Hint: You have given an integer array and a number, you need to write a program to find
+all elements in the array whose sum is equal to the given number. Remember, the array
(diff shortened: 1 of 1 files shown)


src/madhuraMJun20/scannerClass/Exam3.java
```diff
@@ -0,0 +1,38 @@
...
+(2, 5)
+(4, 3)
+(5, 2)
+(3, 4)
+(8, -1)
+(-1, 8)*/
+public class Exam3 {
+	void findPairs(int[] num, int givenSum) {
+		
+		for (int i = 0; i <= num.length; i++) {
```
harshvegada: This code will break on last point at runtime
please do < instead of <= on both places.
Other logic is correct!

harshvegada (commented):
Please do changes again

src/madhuraMJun20/scannerClass/Exam3.java
MadhuraMJun20: Changes done.

harshvegada (approved):
Good job @MadhuraMJun20 
From next time take of this arrayIndexOutOfBoundExceotion
Logic is correct!
Need more practice

PR #572, opened 2020-06-28
Vaishnavi_Assignment_20_60Min_Done

VaishnaviVJun20:
image
image
image
image

### DIFF
--- a/src/vaishnaviVJun20_arrayExamples/Armstrong.java
+++ b/src/vaishnaviVJun20_arrayExamples/Armstrong.java
@@ -1,0 +1,27 @@
+package vaishnaviVJun20_arrayExamples;
+
+//WAP to check if a number is Armstrong or not.
+
+public class Armstrong {
+	
+	void findIfNumberIsArmstrong(int input) {
+		int Originalnumber = input;
+		int sum =0;
+		while(input>0) {
+			int num = input%10;
+			input = input/10;
+			int result = num*num*num;
+			sum += result;
+		}	
+		if(sum == Originalnumber)
+			System.out.println(Originalnumber +" is a armstrong number");
+		else
+			System.out.println(Originalnumber +" is not a armstrong number");
+	}
+
+	public static void main(String[] args) {
+		Armstrong armstrong = new Armstrong();
+		int input = 153;
+		armstrong.findIfNumberIsArmstrong(input);
+	}
+}
--- a/src/vaishnaviVJun20_arrayExamples/Armstrong.java
+++ b/src/vaishnaviVJun20_arrayExamples/Armstrong.java
@@ -8,10 +8,8 @@ public class Armstrong {
 		int Originalnumber = input;
 		int sum =0;
 		while(input>0) {
-			int num = input%10;
+			sum += Math.pow((input % 10), 3);
 			input = input/10;
-			int result = num*num*num;
-			sum += result;
 		}	
 		if(sum == Originalnumber)
 			System.out.println(Originalnumber +" is a armstrong number");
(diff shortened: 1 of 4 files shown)


src/vaishnaviVJun20_arrayExamples/Armstrong.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class Armstrong {
+	
+	void findIfNumberIsArmstrong(int input) {
+		int Originalnumber = input;
+		int sum =0;
+		while(input>0) {
+			int num = input%10;
+			input = input/10;
+			int result = num*num*num;
```
AditiGJUN20: **Suggestion:**
You can take a look at power function - Math.pow(x,y)
which will help you achieve same result for multiplication of same number multiple times (to find square, cube and so on).
i.e. Math.pow((number % 10), 3))

src/vaishnaviVJun20_arrayExamples/Armstrong.java
```diff
@@ -0,0 +1,27 @@
...
+			sum += result;
```
AditiGJUN20: On this line you can directly formulate the value of variable sum.
No need to declare extra variables and store values separately.(remove line 11 and 13)

src/vaishnaviVJun20_arrayExamples/ReverseInteger.java
```diff
@@ -0,0 +1,21 @@
...
+//WAP to revere number (Consider Input as Integer not as String)
+
+public class ReverseInteger {
+	
+	void reverseIntegerNumber(int num) {
+		System.out.print("Reverse number is:");
+		while(num>0) {
+			int reverse = num%10;
+			num = num/10;
+			System.out.print(reverse);
```
AditiGJUN20: Here you are directly printing the reverse order of the given number, 
Implement integer reverse logic - which is addition of one more statement

AditiGJUN20 (changes requested):

src/vaishnaviVJun20_arrayExamples/Armstrong.java
VaishnaviVJun20: --Implemented this logic using this function
--Thanks for the suggestion

src/vaishnaviVJun20_arrayExamples/Armstrong.java
VaishnaviVJun20: --Done

src/vaishnaviVJun20_arrayExamples/ReverseInteger.java
VaishnaviVJun20: --suggested changes done

AditiGJUN20 (approved):
Good job @VaishnaviVJun20 
Changes are made as per comments.

src/vaishnaviVJun20_arrayExamples/Armstrong.java
```diff
@@ -0,0 +1,25 @@
+package vaishnaviVJun20_arrayExamples;
+
+//WAP to check if a number is Armstrong or not.
+
+public class Armstrong {
+	
+	void findIfNumberIsArmstrong(int input) {
+		int Originalnumber = input;
```
KrishnaKTechnocredits, line 8: Variable name started with capital letter !!!

src/vaishnaviVJun20_arrayExamples/DuplicateNumbers.java
```diff
@@ -0,0 +1,22 @@
+package vaishnaviVJun20_arrayExamples;
+
+//Find duplicate number on Integer array in Java?
+
+public class DuplicateNumbers {
+	
+	void findDuplicateNumber(int[] array) {
+		for(int index=0;index<array.length;index++) {
+			for(int i = index+1;i<array.length;i++) {
```
KrishnaKTechnocredits, line 9: Instead of i use innerIndex.

PR #574, opened 2020-06-28
Archana_Assignment_20_10min+10min+15min+10min

ArchanaNJun20:
1.MissingInteger 10 min
image
2.DuplicateNumber 10 min
image
3.ReversNumber 15 min
image
4.ArmstrongNumber(checked for 153 and 154 numebrs) 10 min
image
image

### DIFF
--- a/src/archanaNJun20/ArmstrongNumber.java
+++ b/src/archanaNJun20/ArmstrongNumber.java
@@ -1,0 +1,22 @@
+package archanaNJun20;
+
+public class ArmstrongNumber {
+	static void printArmstrongNumber(int number){
+		int oNumber=number;
+		int lastdigit=0;
+		int cubeAddition=0;
+		while(number>0){
+			lastdigit=number%10;
+			number=number/10;
+			cubeAddition=cubeAddition+(lastdigit*lastdigit*lastdigit);
+		}
+		if(oNumber==cubeAddition)
+			System.out.println(oNumber+" Number is Armstrong Number ");
+		else
+			System.out.println(oNumber+ " Number is not Armstrong Number ");
+	}
+	public static void main(String[] args){
+		//int number=153;
+		printArmstrongNumber(154);
+	}
+}
--- a/src/archanaNJun20/DuplicateNumber.java
+++ b/src/archanaNJun20/DuplicateNumber.java
@@ -1,0 +1,17 @@
+package archanaNJun20;
+
+public class DuplicateNumber {
+	void printDuplicateNumber(int array1[]){
+		for(int i=0;i<array1.length;i++) {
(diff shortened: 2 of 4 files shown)


AavrutiDJun20 (commented):
All 4 Programs are looking good to me.

AavrutiDJun20 (approved):

src/archanaNJun20/ArmstrongNumber.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class ArmstrongNumber {
+	static void printArmstrongNumber(int number){
+		int oNumber=number;
+		int lastdigit=0;
+		int cubeAddition=0;
+		while(number>0){
+			lastdigit=number%10;
+			number=number/10;
+			cubeAddition=cubeAddition+(lastdigit*lastdigit*lastdigit);
```
KrishnaKTechnocredits, line 11: Here you should use math function like Math.pow(number%10, 3). 
So no need to create int lastdigit=0; this variable.

src/archanaNJun20/DuplicateNumber.java
```diff
@@ -0,0 +1,17 @@
+package archanaNJun20;
+
+public class DuplicateNumber {
+	void printDuplicateNumber(int array1[]){
+		for(int i=0;i<array1.length;i++) {
+			for(int j=i+1;j<array1.length;j++){
```
KrishnaKTechnocredits, line 6: Better to use index and innerIndex instead of i and j.

src/archanaNJun20/MissingInteger.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class MissingInteger {
+	void printMissingInteger(int array[],int n){
+		int total=((n*(n+1))/2);
+		int sum=0;
+		for(int index=0;index<array.length;index++){
+			sum=sum+array[index];
+		}
+		int differnce=total-sum;
+		System.out.println("Missing number in array is: "+differnce);
```
KrishnaKTechnocredits, line 11: You can directly write total-sum; in syso statement.

src/archanaNJun20/ArmstrongNumber.java
ArchanaNJun20: The code is updated based on comments.

src/archanaNJun20/DuplicateNumber.java
ArchanaNJun20: The code is updated based on comments.

src/archanaNJun20/MissingInteger.java
ArchanaNJun20: The code is updated based on comments.

PR #575, opened 2020-06-28
Barkha_Assignment_20_15+10+10+10 mins

BarkhaPJun20:
Program 1: Find the missing number in integer array of 1 to 10?
Hint: You have given an integer array which contains numbers from 1 to 10 but one number is missing, you need to write a Java program to find that missing number in an array.

Program 2: Find duplicate number on Integer array in Java?
Hint: There is exactly one number is repeated in the array. You need to write a program to
find that duplicate number. For example, if an array with length 6 contains numbers {0, 3,
1, 2, 3}, then duplicated number is 3.

Program 3: WAP to revers number (Consider Input as Integer not as String).
Example:-
Input:- 12345
Output:- 54321

Program 4: WAP to check if a number is Armstrong or not.
Example:-
Input:- 153
Output:-153
hint:- (1*1*1)+(5*5*5)+(3*3*3)
1+125+27
153

image
image
image
image

### DIFF
--- a/src/barkhaPJun20/ArmStrongNum.java
+++ b/src/barkhaPJun20/ArmStrongNum.java
@@ -7,18 +7,19 @@ public class ArmStrongNum {
 		int sum=0;
 		while(number>0) {
 			int lastnum=number%10;           //3
-			int cube=lastnum*lastnum*lastnum; 
-			sum=sum+cube;
-			number=number/10;                //51
-		}	                			
+			sum=sum+(lastnum*lastnum*lastnum);
+			number=number/10;               //51
+		}	              
 		if (sum==originalNumber)
 			System.out.println("Number "+originalNumber+" is an Armstrong number");	          
 		else
-			System.out.println("Number "+originalNumber+" is not an Armstrong number");
+			System.out.println("Number "+originalNumber+" is not an Armstrong number");	                			
 	}	
 	public static void main(String[] args) {
 		ArmStrongNum armStrongNum=new ArmStrongNum();
 		int num=153;
+		int num1=15622;
 		armStrongNum.findArmStrongNum(num);
+		armStrongNum.findArmStrongNum(num1);
 	}
-}
+}	
--- a/src/barkhaPJun20/ArmStrongNum.java
+++ b/src/barkhaPJun20/ArmStrongNum.java
@@ -1,0 +1,24 @@
+package barkhaPJun20;
+
+public class ArmStrongNum {
+	
+	void findArmStrongNum(int originalNumber) {
+		int number=originalNumber;
+		int sum=0;
+		while(number>0) {
+			int lastnum=number%10;           //3
+			int cube=lastnum*lastnum*lastnum; 
+			sum=sum+cube;
+			number=number/10;                //51
+		}	                			
+		if (sum==originalNumber)
+			System.out.println("Number "+originalNumber+" is an Armstrong number");	          
+		else
+			System.out.println("Number "+originalNumber+" is not an Armstrong number");
+	}	
+	public static void main(String[] args) {
+		ArmStrongNum armStrongNum=new ArmStrongNum();
+		int num=153;
+		armStrongNum.findArmStrongNum(num);
+	}
+}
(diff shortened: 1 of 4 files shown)


src/barkhaPJun20/ArmStrongNum.java
```diff
@@ -0,0 +1,24 @@
+package barkhaPJun20;
+
+public class ArmStrongNum {
+	
+	void findArmStrongNum(int originalNumber) {
+		int number=originalNumber;
+		int sum=0;
+		while(number>0) {
+			int lastnum=number%10;           //3
+			int cube=lastnum*lastnum*lastnum;
```
SwapnaWJun20: You can directly use sum to store the value. like
sum += lastname * lastname *  lastname;

src/barkhaPJun20/ReverseInteger.java
```diff
@@ -0,0 +1,21 @@
+package barkhaPJun20;
+
+public class ReverseInteger {
+	
+	void printReverseInt(int originalNumber) {
+		int num=originalNumber;
+		System.out.println("Reverse of Number: "+originalNumber);
+
```
SwapnaWJun20: Please remove extra spaces

src/barkhaPJun20/ReverseInteger.java
```diff
@@ -0,0 +1,21 @@
...
+public class ReverseInteger {
+	
+	void printReverseInt(int originalNumber) {
+		int num=originalNumber;
+		System.out.println("Reverse of Number: "+originalNumber);
+		
+		while(num>0) {
+		int lastDigit=num%10;   //5
+		num=num/10;            //1234
+
```
SwapnaWJun20: Please remove extra spaces

src/barkhaPJun20/FindMissingNumInArray.java
```diff
@@ -0,0 +1,21 @@
+package barkhaPJun20;
+
+public class FindMissingNumInArray {
+	int arraySum=0;
+	int originalSum=0;
+	
+	void printMissingNumInArray(int[] arrayNum) {
+		for (int index=0; index<arrayNum.length; index++) {
+			arraySum=arraySum+arrayNum[index];
```
SwapnaWJun20, line 9: You can write for loop (for array values) inside for loop  (for range specified) i and compare the the two with each other.

BarkhaPJun20: Wrote this method as well. In fact wrote both methods so in future I can refer both.Thank you.

SwapnaWJun20 (approved):

src/barkhaPJun20/FindMissingNumInArray.java
SwapnaWJun20: Code looks good to me. :)

src/barkhaPJun20/ArmStrongNum.java
```diff
@@ -0,0 +1,25 @@
+package barkhaPJun20;
+
+public class ArmStrongNum {
+	
+	void findArmStrongNum(int originalNumber) {
+		int number=originalNumber;
+		int sum=0;
+		while(number>0) {
+			int lastnum=number%10;           //3
+			sum=sum+(lastnum*lastnum*lastnum);
```
KrishnaKTechnocredits, line 10: Use math function here, like Math.pow(number%10, 3);
And never initialize variable in loops. like int lastnum = ; this.

src/barkhaPJun20/FindMissingNumInArray.java
```diff
@@ -0,0 +1,35 @@
...
+public class FindMissingNumInArray {
+	int arraySum=0;
+	int originalSum=0;
+	//Method 1:Find Missing number by find difference in addition.
+	void printMissingNumInArray(int[] arrayNum) {
+		for (int index=0; index<arrayNum.length; index++) {
+			arraySum=arraySum+arrayNum[index];
+		}
+		for (int num=1; num<=10; num++) {
+			originalSum=originalSum+num;
```
KrishnaKTechnocredits, line 12: Here you can write n(n+1)/2 formula as well.
void findMissingNumber2(int[] array) {
		int al = array.length+1;
		int expectedSum = (al*(al+1))/2;
		int actualSum = 0;
		for(int index=0; index<al-1; index++) {
			actualSum = actualSum + array[index];
		}
		System.out.println("expectedSum "+expectedSum);
		System.out.println("Missing number in given array is: "+(expectedSum-actualSum));
	}

PR #579, opened 2020-06-28
Aastha_Assignment_20_Done_40mins

AasthaJJun20:
->NumbersAndArrays.java is pushed for Assignment-20
->PFB screenshot:
NumbersAndArrays-output
Kindly Review,
Thanks

### DIFF
--- a/src/aasthaJJun20/NumbersAndArrays.java
+++ b/src/aasthaJJun20/NumbersAndArrays.java
@@ -1,0 +1,77 @@
+package aasthaJJun20;
+
+public class NumbersAndArrays {
+	//This method is written just to display the array elements in output
+	void displayArray(int[] input) {
+		System.out.print("Input Array : { ");
+		for(int index = 0; index < input.length; index++) {
+			System.out.print(input[index]+" ");;
+		}
+		System.out.println("}");
+	}
+	
+	// Program 1: Find the missing number in integer array of 1 to 10?
+	void findMissingNumberFromArray(int[] input) {
+		displayArray(input);
+		int actualSum = 0, differedSum = 0;
+		for (int index = 1; index <= 10; index++) {
+			actualSum += index;
+		}
+		for (int index = 0; index < input.length; index++) {
+			differedSum += input[index];
+		}
+		System.out.println("The missing integer is: " + (actualSum - differedSum));
+
+	}
+
+	// Program 2: Find duplicate number on Integer array in Java? when exactly one number is repeated in the array
+	void findDuplicateInArray(int[] input) {
+		displayArray(input);
+		for (int index = 0; index < input.length; index++) {
+			for (int innerIndex = 1; innerIndex < input.length; innerIndex++) {
+				if (input[index] == input[innerIndex] && index != innerIndex) {
+					System.out.println("Duplicate number is: " + input[index]);
+					return;
+				}
+			}
+		}
+	}
+
+	// Program 3: WAP to revers number (Consider Input as Integer not as String).
+	void displayReverseInteger(int num) {
+		System.out.print("Reverse of given number " + num + " is: ");
+		while (num > 0) {
+			System.out.print(num % 10);
+			num /= 10;
+		}
+	}
+
+	// Program 4: WAP to check if a number is Armstrong or not. Note: It can only handle a 3 digit number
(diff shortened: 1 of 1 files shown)


src/aasthaJJun20/NumbersAndArrays.java
```diff
@@ -0,0 +1,77 @@
...
+
+	}
+
+	// Program 2: Find duplicate number on Integer array in Java? when exactly one number is repeated in the array
+	void findDuplicateInArray(int[] input) {
+		displayArray(input);
+		for (int index = 0; index < input.length; index++) {
+			for (int innerIndex = 1; innerIndex < input.length; innerIndex++) {
+				if (input[index] == input[innerIndex] && index != innerIndex) {
+					System.out.println("Duplicate number is: " + input[index]);
```
NikhilMJun20, line 33: This double for loop is working as expected. But it could be written in a little better way and with that the if() condition will also be much cleaner. PFB:-

for (int index = 0; index < input.length; index++) {
		for (int innerIndex = index+1; innerIndex < input.length; innerIndex++) {
				if (input[index] == input[innerIndex] ) {}
              }
}

src/aasthaJJun20/NumbersAndArrays.java
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	// Program 3: WAP to revers number (Consider Input as Integer not as String).
+	void displayReverseInteger(int num) {
+		System.out.print("Reverse of given number " + num + " is: ");
+		while (num > 0) {
+			System.out.print(num % 10);
+			num /= 10;
+		}
```
NikhilMJun20, line 47: The output is exactly as expected.
But, here integer is not reversed exactly - it is chopped digit by digit and printed in reverse order.
Please implement integer reverse logic - which is just a small extension to your login.

src/aasthaJJun20/NumbersAndArrays.java
```diff
@@ -0,0 +1,77 @@
...
+		numbersAndArrays.findMissingNumberFromArray(arr1);
+		System.out.println("\nProgram:2 Output:---------------------------------");
+		numbersAndArrays.findDuplicateInArray(arr2);
+		System.out.println("\nProgram:3 Output:---------------------------------");
+		numbersAndArrays.displayReverseInteger(12345);
+		System.out.println("\n\nProgram:4 Output:---------------------------------");
+		numbersAndArrays.isArmstrong(153);
+		numbersAndArrays.isArmstrong(200);
+	}
+}
```
NikhilMJun20, line 79: Very nice, crisp and clean coding overall!
:-)

src/aasthaJJun20/NumbersAndArrays.java
AasthaJJun20: Right, this logic looks more simplified, will change it. Thanks

src/aasthaJJun20/NumbersAndArrays.java
AasthaJJun20: got it, will implement it. Thanks Nikhil for review.

NikhilMJun20 (approved):

src/aasthaJJun20/NumbersAndArrays.java
```diff
@@ -0,0 +1,79 @@
...
+		}
+		System.out.println("}");
+	}
+	
+	// Program 1: Find the missing number in integer array of 1 to 10?
+	void findMissingNumberFromArray(int[] input) {
+		displayArray(input);
+		int actualSum = 0, differedSum = 0;
+		for (int index = 1; index <= 10; index++) {
+			actualSum += index;
```
KrishnaKTechnocredits, line 18: @AasthaJJun20  @NikhilMJun20  ,To sum of first N number, we can use formula to impress interviewer.

n*(n+1)/2 --> (10*11)/2 = 55 :)

KrishnaKTechnocredits (approved):
Code looks good to me.

src/aasthaJJun20/NumbersAndArrays.java
AasthaJJun20: Yes Krishna, this formula will make the logic simple and short as well. Thanks

PR #585, opened 2020-06-29
AnupS_Assignment_20_Done_50mins

AnupSJun20:
AnupS_Assignment_20_Done_50mins_Please review.
Program1
Program2
Program3
Program4

### DIFF
--- a/src/anupSJun20/ArmstrongNumber.java
+++ b/src/anupSJun20/ArmstrongNumber.java
@@ -1,0 +1,28 @@
+package anupSJun20;
+/*Program 4: WAP to check if a number is Armstrong or not.
+Example:-
+Input:- 153
+Output:-153
+hint:- (1*1*1)+(5*5*5)+(3*3*3)
+1+125+27
+153*/
+
+public class ArmstrongNumber {
+	static void findANumberArmstrong(int num) {
+		int input = num;
+		int newNum = 0;
+		while(num>0) {
+			newNum = newNum+ (num%10)*(num%10)*(num%10);
+			num = num/10;
+		}
+		if(newNum ==input) 
+			System.out.println("The number " + input + " is an armstrong Number");
+		else
+			System.out.println("The number " + input+ " is not an armstrong Number");
+		
+	}
+	public static void main(String[] args) {
+		findANumberArmstrong(153);
+	}
+
+}
--- a/src/anupSJun20/ArmstrongNumber.java
+++ b/src/anupSJun20/ArmstrongNumber.java
@@ -10,19 +10,17 @@ hint:- (1*1*1)+(5*5*5)+(3*3*3)
 public class ArmstrongNumber {
 	static void findANumberArmstrong(int num) {
 		int input = num;
-		int newNum = 0;
+		double newNum = 0;
 		while(num>0) {
-			newNum = newNum+ (num%10)*(num%10)*(num%10);
+			newNum = newNum + Math.pow((num%10), 3);
 			num = num/10;
 		}
 		if(newNum ==input) 
 			System.out.println("The number " + input + " is an armstrong Number");
 		else
-			System.out.println("The number " + input+ " is not an armstrong Number");
-		
+			System.out.println("The number " + input+ " is not an armstrong Number");	
 	}
 	public static void main(String[] args) {
 		findANumberArmstrong(153);
 	}
-
 }
--- a/src/anupSJun20/MissingNumbrinIntegerArray.java
+++ b/src/anupSJun20/MissingNumbrinIntegerArray.java
@@ -11,9 +11,8 @@ public class MissingNumbrinIntegerArray {
 			sumofArray = sumofArray+num[index];
 		}
 		return overallSum - sumofArray;
-		
-		
 	}
+	
 	public static void main(String[] args) {
 		int[] num = {1,2,3,4,5,7,10,9,8};
 		System.out.println("The missing number is : " + findTheMissingNumber(num,1,10));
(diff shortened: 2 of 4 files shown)


src/anupSJun20/ArmstrongNumber.java
```diff
@@ -0,0 +1,28 @@
...
+hint:- (1*1*1)+(5*5*5)+(3*3*3)
+1+125+27
+153*/
+
+public class ArmstrongNumber {
+	static void findANumberArmstrong(int num) {
+		int input = num;
+		int newNum = 0;
+		while(num>0) {
+			newNum = newNum+ (num%10)*(num%10)*(num%10);
```
AditiGJUN20: **Suggestion:**
You can take a look at power function - Math.pow(x,y)
which will help you achieve same result for multiplication.
i.e. Math.pow((number % 10), 3))

src/anupSJun20/MissingNumbrinIntegerArray.java
```diff
@@ -0,0 +1,21 @@
...
+		for(int index = 0;index<num.length;index++) {
+			sumofArray = sumofArray+num[index];
+		}
+		return overallSum - sumofArray;
+		
+		
+	}
+	public static void main(String[] args) {
+		int[] num = {1,2,3,4,5,7,10,9,8};
+		System.out.println("The missing number is : " + findTheMissingNumber(num,1,10));
```
AditiGJUN20, line 18: @AnupSJun20 
Have you tried to make code generic by passing extra arguments in method?
If yes then - 
for ->int[] num = {11,15,12,13} when i call findTheMissingNumber(num,11,15) it's giving me incorrect result.

src/anupSJun20/ReverseNumber.java
```diff
@@ -0,0 +1,24 @@
...
+/* Program 3: WAP to reverse number (Consider Input as Integer, not as String).
+Example:-
+Input:- 12345
+Output:- 54321*/
+
+public class ReverseNumber {
+	static void findReverseNumber(int num) {
+		int reverseNum = 0;
+		while(num > 0) {
+			reverseNum = reverseNum*10;
```
AditiGJUN20: - You can reduce the statement by following syntax:
reverseNum=reverseNum*10+(inputNum%10);

src/anupSJun20/MissingNumbrinIntegerArray.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class MissingNumbrinIntegerArray {
+	static int findTheMissingNumber(int[] num,int startNum,int endNum) {
+		int overallSum = (endNum*(endNum+1))/2;
+		int sumofArray = 0;
+		for(int index = 0;index<num.length;index++) {
+			sumofArray = sumofArray+num[index];
+		}
+		return overallSum - sumofArray;
+
```
AditiGJUN20: remove extra lines

AditiGJUN20 (changes requested):
- I've added separate comment where changes are required.
- Overall for all program remove extra blank lines.
i.e. for MissingNumbrinIntegerArray.java i've added comment.

src/anupSJun20/ArmstrongNumber.java
AnupSJun20: Thanks Aditi. Changes incorporated. Didn't think of this function.

src/anupSJun20/MissingNumbrinIntegerArray.java
AnupSJun20: Thanks Aditi for this comment. I believe Maulik said that try and take a hard coded range 1- 10 as of now and he will explain in the class for user defined values and multiple missing numbers. So once Maulik will explain today, will create a new program. 
Yes the code gave you incorrect result as it was created for a hard code value to start from 1.

src/anupSJun20/ReverseNumber.java
AnupSJun20: Thanks for the comment. Incorporated the changes and pushed.

src/anupSJun20/MissingNumbrinIntegerArray.java
AnupSJun20: Removed now.

src/anupSJun20/ArmstrongNumber.java
```diff
@@ -0,0 +1,26 @@
...
+hint:- (1*1*1)+(5*5*5)+(3*3*3)
+1+125+27
+153*/
+
+public class ArmstrongNumber {
+	static void findANumberArmstrong(int num) {
+		int input = num;
+		double newNum = 0;
+		while(num>0) {
+			newNum = newNum + Math.pow((num%10), 3);
```
AditiGJUN20, line 15: There is no need to change data type of newNum from int to double due to power function.
you can use typecasting here.
newNum = (int) (newNum + (Math.pow((number % 10), 3)));

src/anupSJun20/MissingNumbrinIntegerArray.java
```diff
@@ -0,0 +1,20 @@
+package anupSJun20;
+/*Program 1: Find the missing number in integer array of 1 to 10?
+Hint: You have given an integer array which contains numbers from 1 to 10 but one number is missing, 
+you need to write a Java program to find that missing number in an array.*/
+
+public class MissingNumbrinIntegerArray {
+	static int findTheMissingNumber(int[] num,int startNum,int endNum) {
```
AditiGJUN20, line 7: **// Suggestion -**
Here, in entire method you are not using startNum integer(neither in calculation or in any SOP statements). So there is no need to pass it in method argument and you can reduce the argument as **static int findTheMissingNumber(int[] num, int endNum)**

Another way is, just pass input array in method argument- **static int findTheMissingNumber(int[] num)** and you can use **int endNum = array.length+1**

src/anupSJun20/MissingNumbrinIntegerArray.java
AditiGJUN20: Yes Anup, this will work for hardcoded value 1-10.

AditiGJUN20 (approved):
Added suggestion for future code reference.
Rest Code looks good. 
Approved for further review and merge.

PR #586, opened 2020-06-29
Rachana_Assignment_20_Done_45min

RachanaGJun20:
m20
-----------------------------------------------------------------------------------------
duplicate20
-----------------------------------------------------------------------------------------
reverse20e
-----------------------------------------------------------------------------------------
armstrong20

### DIFF
--- a/src/rachanaGJun20/Array/ArmstrongNumber.java
+++ b/src/rachanaGJun20/Array/ArmstrongNumber.java
@@ -4,16 +4,16 @@ import java.util.Scanner;
 
 public class ArmstrongNumber {
 
-	//method will find whether given number is Armstrong 
+	// method will find whether given number is Armstrong
 	boolean isArmstrongNumber(int number) {
 		int remainder = 0;
-		int sum = 0;
+		double sum = 0;
 		int givenNumber = number;
 
 		while (number != 0) {
 			remainder = number % 10;
 			number = number / 10;
-			sum = sum + (remainder * remainder * remainder);
+			sum = sum + Math.pow(remainder, 3);
 		}
 		return (givenNumber == sum);
 	}
--- a/src/rachanaGJun20/Array/ArmstrongNumber.java
+++ b/src/rachanaGJun20/Array/ArmstrongNumber.java
@@ -1,0 +1,35 @@
+package rachanaGJun20.Array;
+
+import java.util.Scanner;
+
+public class ArmstrongNumber {
+
+	//method will find whether given number is Armstrong 
+	boolean isArmstrongNumber(int number) {
+		int remainder = 0;
+		int sum = 0;
+		int givenNumber = number;
+
+		while (number != 0) {
+			remainder = number % 10;
+			number = number / 10;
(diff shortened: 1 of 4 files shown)


src/rachanaGJun20/Array/ArmstrongNumber.java
```diff
@@ -0,0 +1,35 @@
...
+	//method will find whether given number is Armstrong 
+	boolean isArmstrongNumber(int number) {
+		int remainder = 0;
+		int sum = 0;
+		int givenNumber = number;
+
+		while (number != 0) {
+			remainder = number % 10;
+			number = number / 10;
+			sum = sum + (remainder * remainder * remainder);
```
AjitSJune20: Suggestion: you can use  Math.pow function  as well
Number of Variables can be reduced.

src/rachanaGJun20/Array/DuplicateNumberCheck.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class DuplicateNumberCheck{
+
+	// method will find and display duplicate number from given array
+	void findDuplicateNumber(int[] arr) {
+		for (int i = 0; i < arr.length; i++) {
+			for (int j = i + 1; j < arr.length; j++) {
+				if (arr[i] == arr[j]) {
+					System.out.println("Duplicate number in a given array is: " + arr[i]);
+					break;
```
AjitSJune20, line 18: Please add else statement to handle scenario If there is no duplicate element 
Example : arr = { 0, 3, 1, 2}
We wont get any result if no duplicate element found

src/rachanaGJun20/Array/ArmstrongNumber.java
RachanaGJun20: Sure will Update.. Thank you :)

src/rachanaGJun20/Array/DuplicateNumberCheck.java
RachanaGJun20: yup got it.... Will update Thanks :)

src/rachanaGJun20/Array/MissingNumberInArray.java
```diff
@@ -0,0 +1,34 @@
...
+		return sum;
+	}
+
+	// method will find missing number between 1 to 10
+	void findMissingNumberInArray(int[] arr) {
+		int sumOfGivenArray = 0;
+		for (int i = 0; i < arr.length; i++) {
+			sumOfGivenArray = sumOfGivenArray + arr[i];
+		}
+		System.out.println("Missing number is: " + (SumOf1To10() - sumOfGivenArray));
```
AjitSJune20, line 23: Looks good . Just one input 
just handle this scenario if input is { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 } then it will give answer "Missing number is 0" , which should not be the case.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #587, opened 2020-06-29
Swapna_Assignment_20_Done_30+10+10+10_min

SwapnaWJun20:
image
image
image
image

### DIFF
--- a/src/swapnaWJun20/Armstrong.java
+++ b/src/swapnaWJun20/Armstrong.java
@@ -1,0 +1,23 @@
+package swapnaWJun20;
+
+public class Armstrong {
+
+	//Armstrong number logic
+	void checkValueIfArmstrong(int num) {
+		int num1 = num;
+		int sum = 0;
+		while (num1 > 0) {
+			sum += (num1 % 10) * (num1 % 10) * (num1 % 10);
+			num1 = num1 / 10;
+		}
+		if (num == sum)
+			System.out.println(num + " is Armstrong");
+		else
+			System.out.println(num + " is not Armstrong");
+	}
+
+	public static void main(String[] args) {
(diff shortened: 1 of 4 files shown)


AnupSJun20 (approved):
Code looks good to me.

src/swapnaWJun20/ReverseNo.java
```diff
@@ -0,0 +1,20 @@
...
+public class ReverseNo {
+
+	//Reversing the number
+	void doReverse(int inp) {
+		int revNo = 0;
+		while (inp > 0) {
+			revNo = revNo * 10;
+			revNo = revNo + inp % 10;
+			inp = inp / 10;
+		}
```
KrishnaKTechnocredits, line 12: But, here integer is not reversed exactly - it is chopped digit by digit and printed in reverse order.
Please implement integer reverse logic - which is just a small extension to your login.
void findReverseNumber(int number) {
		int reverseNum = 0;
		int mod =0;
		while(number>0) {
			mod = number%10;
			number = number/10;
		    reverseNum = mod + reverseNum*10 ;
		}
		System.out.println("Reverse Number is "+reverseNum);
	}

src/swapnaWJun20/Armstrong.java
```diff
@@ -0,0 +1,23 @@
+package swapnaWJun20;
+
+public class Armstrong {
+
+	//Armstrong number logic
+	void checkValueIfArmstrong(int num) {
+		int num1 = num;
+		int sum = 0;
+		while (num1 > 0) {
+			sum += (num1 % 10) * (num1 % 10) * (num1 % 10);
```
KrishnaKTechnocredits, line 10: Use Math.pow(num1 % 10), 3) here.

PR #591, opened 2020-06-29
ADT_Assignment20_Done_45Mins

AditiGJUN20:
### Assignment 20:

**1. MissingNumberInIntArray.java**
Time: 20Mins
Screen Shot 2020-06-28 at 11 08 52 PM
**Note:**
- Code run for any numbered series where exactly one number is missing. I tried to set flag/condition for the case where there is no missing element but seems i'm missing something/failing to achieve that. I'll try to get that done. If not then will ask Maulik.

**2. DuplicateNumInArray.java**
Time: 6 Mins
Screen Shot 2020-06-28 at 11 10 25 PM

**3. ReverseNumber.java**
Time: 5 Mins
Screen Shot 2020-06-28 at 11 10 38 PM

**4. ArmstrongNumber.java**
Time: 10mins
Screen Shot 2020-06-28 at 11 10 54 PM

### DIFF
--- a/src/aditiGJUN20/MissingNumberInIntArray.java
+++ b/src/aditiGJUN20/MissingNumberInIntArray.java
@@ -1,0 +1,35 @@
+/*
+Program 1: Find the missing number in integer array of 1 to 10?
+Hint: You have given an integer array which contains numbers from 1 to 10 
+but one number is missing, you need to write a Java program to find that 
+missing number in an array
+*/
+
+package aditiGJUN20;
+
+import java.util.Arrays;
+
+public class MissingNumberInIntArray {
+	void findMissingNumber(int[] intArray) {
+		Arrays.sort(intArray);
+		int arrayValue = intArray[0];
+		for (int index = 0; index < intArray.length; index++) {
+			while (intArray[index] == arrayValue)
+				arrayValue++;
+		}
+		System.out.println("Missing number is: " + arrayValue);
+	}
+
+	public static void main(String[] args) {
+		MissingNumberInIntArray missingNumberInIntArray = new MissingNumberInIntArray();
+		// input- 1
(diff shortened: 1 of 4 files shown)


AmitaRJun20 (approved):
codes are nicely written . Math functionality and Array.sort method you have used is impressive.

src/aditiGJUN20/MissingNumberInIntArray.java
```diff
@@ -0,0 +1,36 @@
...
+			while (intArray[index] == arrayValue)
+				arrayValue++;
+		}
+		System.out.println("Missing number is: " + arrayValue);
+	}
+
+	public static void main(String[] args) {
+		MissingNumberInIntArray missingNumberInIntArray = new MissingNumberInIntArray();
+		// input- 1
+		int[] inputarray1 = { 3, 4, 2, 7, 5, 6, 8 };
```
KrishnaKTechnocredits, line 26: Number 1 is missing here isn't it ? output shows number 9 is missing. 

Reason : we sort an array and we considered first number as starting point, in array 
int[] inputarray1 = { 3, 4, 2, 7, 5, 6, 8 } , number 2 is on first index, it will never talk about 1.  please correct me if I misinterpreted your logic .

If you want to find missing number starting from lowest number in given array then your logic is correct.

KrishnaKTechnocredits (approved):
Code looks good to me. Please take a look at my comments.

src/aditiGJUN20/MissingNumberInIntArray.java
AditiGJUN20: Yes Krishna. Here, it's taking array input's lowest number as starting point. 
I'm going to rewrite this code and will push that logic.
Thank you for the feedback.

PR #593, opened 2020-06-29
Kartikey_Assignment_20_Done_90+10+10+15min

KartikeyDJune20:
image
image
image
image

### DIFF
--- a/src/kartikeyDJune20/ArmstrongNumber.java
+++ b/src/kartikeyDJune20/ArmstrongNumber.java
@@ -1,0 +1,22 @@
+package kartikeyDJune20;
+
+public class ArmstrongNumber {
+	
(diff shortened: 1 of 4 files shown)


src/kartikeyDJune20/ArmstrongNumber.java
```diff
@@ -0,0 +1,22 @@
+package kartikeyDJune20;
+
+public class ArmstrongNumber {
+	
+	static void armstrongNum(int num) {
+	int originalNum=num;
+		int digit=0;
+		do {
+		digit=digit+((num%10)*(num%10)*(num%10));
```
KrishnaKTechnocredits, line 9: You can use Math.pow(num%10, 3) here.

PR #594, opened 2020-06-29
Piyush_Assignment_20_1.5Hour

PiyushPJun20:
[Screenshot.docx](https://github.com/KrishnaKTechnocredits/JAVATechnoJun20/files/4844897/Screenshot.docx)

### DIFF
--- a/src/Armstrong.java
+++ b/src/Armstrong.java
@@ -1,0 +1,22 @@
+
+public class Armstrong {
+	void ArmstrongNum(int num) {
+
+		int aNum = 0, temp = num;
+		while (temp > 0) {
+			aNum += ((temp % 10) * (temp % 10) * (temp % 10));
+			temp = temp / 10;
+		}
+		if (num == aNum) {
+			System.out.println(num + " is Armstrong Number");
+		} else {
+			System.out.println(num + " is Not Armstrong Number");
+		}
+	}
+
+	public static void main(String[] args) {
+		Armstrong armstrong = new Armstrong();
+		int num = 153;
+		armstrong.ArmstrongNum(num);
+	}
+}
--- a/src/DuplicateNum.java
+++ b/src/DuplicateNum.java
@@ -1,0 +1,21 @@
+public class DuplicateNum {
+
+	void displayDuplicate(int[] intArr) {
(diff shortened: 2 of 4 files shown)


src/DuplicateNum.java
```diff
@@ -0,0 +1,21 @@
+public class DuplicateNum {
+
+	void displayDuplicate(int[] intArr) {
+
+		System.out.print(" Duplicated Number is : ");
+		for (int i = 0; i < intArr.length; i++) {
+			for (int j = i + 1; j < intArr.length; j++) {
+				if (intArr[i] == intArr[j]) {
+					System.out.println(intArr[i]);
+
```
ShrutiDJun20, line 10: Just a suggestion here: we should also handle the scenario where what should happen if there is no duplicate in the array. But no need to change it in the code as your code is as per requirement..

ShrutiDJun20 (approved):

src/DuplicateNum.java
KrishnaKTechnocredits: @ShrutiDJun20 you are correct ! Ideally we should include that.

src/Armstrong.java
```diff
@@ -0,0 +1,22 @@
+
+public class Armstrong {
+	void ArmstrongNum(int num) {
+
+		int aNum = 0, temp = num;
+		while (temp > 0) {
+			aNum += ((temp % 10) * (temp % 10) * (temp % 10));
```
KrishnaKTechnocredits, line 7: as an alternative, we can use Math.pow(temp%10,3) function when

src/MissingNumber.java
```diff
@@ -0,0 +1,28 @@
+public class MissingNumber {
+
+	void displayMissingNum(int[] intArr) {
+		
+		boolean flag;
+		System.out.println(" Missing Numbers are : ");
+		for (int i=1; i<=10; i++) {
```
KrishnaKTechnocredits, line 7: Please use index instead of i, it will be good during interview as well as in regular practice.

KrishnaKTechnocredits (approved):
Code looks good to me. Keep it up.

PR #597, opened 2020-06-29
Ajit_Assignment_20 80 mins

AjitSJune20:
#1 
image

#2
image

#3 
image

#4 
image

### DIFF
--- a/src/ajitSJune20/FindDuplicateNumInIntAray.java
+++ b/src/ajitSJune20/FindDuplicateNumInIntAray.java
@@ -1,0 +1,31 @@
+//Find duplicate number on Integer array in Java?  Example : {0, 3,1, 2, 3}, then duplicated number is 3.
+package ajitSJune20;
+
+public class FindDuplicateNumInIntAray {
+
+	int getDuplicateNumInArray(int[] array) {
+		int dupNum = -1;
+		for (int index1 = 0; index1 < array.length; index1++) {
+			for (int index2 = (index1 + 1); index2 < array.length; index2++) {
+				if (array[index1] == array[index2]) {
+					dupNum = array[index2];
+					break;
+				}
+			}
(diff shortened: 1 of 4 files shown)


src/ajitSJune20/FindDuplicateNumInIntAray.java
```diff
@@ -0,0 +1,31 @@
...
+package ajitSJune20;
+
+public class FindDuplicateNumInIntAray {
+
+	int getDuplicateNumInArray(int[] array) {
+		int dupNum = -1;
+		for (int index1 = 0; index1 < array.length; index1++) {
+			for (int index2 = (index1 + 1); index2 < array.length; index2++) {
+				if (array[index1] == array[index2]) {
+					dupNum = array[index2];
```
KartikeyDJune20, line 11: We could have simplified it.. the moment if condition is true that means Duplicate no. is there so Print that time only.. and what if the duplicate no. is -1.. instead of -1 we can use boolean

src/ajitSJune20/FindReverseOfNumber.java
```diff
@@ -0,0 +1,20 @@
...
+		do {
+			reverse = reverse * 10 + (number % 10);
+			number = number / 10;
+
+		} while (number > 0);
+		return reverse;
+	}
+
+	public static void main(String[] args) {
+
```
KartikeyDJune20, line 16: We can minimize extra spacing.. rest is okay

src/ajitSJune20/FindDuplicateNumInIntAray.java
AjitSJune20: If there is no  duplicate number found then It will print "No duplicate element found in Array"
image

I agree we can use Boolean as well/

PR #598, opened 2020-06-29
Raj_Assignment_20_80min_Done

RajAJun20:
Assignment_20_P1
Assignment_20_P2
Assignment_20_P3
Assignment_20_P4

### DIFF
--- a/src/rajAJun20/ArmstrongNumber.java
+++ b/src/rajAJun20/ArmstrongNumber.java
@@ -1,0 +1,27 @@
+package rajAJun20;
+
+public class ArmstrongNumber {
+	
+	void findIsNumberArmstrong(int input){
+		int actualNumber = input ;
+		int sum = 0;
+		while(input > 0) {
+			int num = input % 10 ;
+			input = input / 10;
+			int result = num*num*num ;
+			sum = sum + result; 
+	}
+		if(sum == actualNumber) {
+			System.out.println(actualNumber + " is an Armstrong number");
+		} else {
+			System.out.println(actualNumber + " is not an Armstrong number");
+		}
+	}
+
+		public static void main(String[] args) {
+			ArmstrongNumber armstrongNumber = new ArmstrongNumber();
+			
+			int number = 153;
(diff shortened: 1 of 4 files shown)


src/rajAJun20/ArmstrongNumber.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class ArmstrongNumber {
+	
+	void findIsNumberArmstrong(int input){
+		int actualNumber = input ;
+		int sum = 0;
+		while(input > 0) {
+			int num = input % 10 ;
+			input = input / 10;
+			int result = num*num*num ;
```
KrishnaKTechnocredits, line 11: Never initialized variable inside loops. 
No need to use result variable.

src/rajAJun20/MissingNumber.java
```diff
@@ -0,0 +1,23 @@
+package rajAJun20;
+
+public class MissingNumber {
+	
+	void findMissingNumber(int[] input) {
+		int sum = 0, differSum = 0;
+		for (int index = 1; index <= 10; index++) {
```
KrishnaKTechnocredits, line 7: You can use formula like n*(n+1)/2 . This way you can avoid for loop.

src/rajAJun20/ReverseNumber.java
```diff
@@ -0,0 +1,20 @@
+package rajAJun20;
+
+public class ReverseNumber {
+	
+	void printReverseNumber(int inputNumber) {
+		while(inputNumber > 0) {
+			int reverse = inputNumber % 10;
+			inputNumber = inputNumber/10;
+			System.out.println("Reverse number is : " + reverse);
```
KrishnaKTechnocredits, line 9: This will not work. 
void findReverseNumber(int number) {
		int reverseNum = 0;
		int mod =0;
		while(number>0) {
			mod = number%10;
			number = number/10;
		    reverseNum = mod + reverseNum*10 ;
		}
		System.out.println("Reverse Number is "+reverseNum);
	}

PR #607, opened 2020-06-30
Nikhil_Assignment_21_Done_(3+10=)13Min

NikhilMJun20:
->Completed and pushed Assignment # 21
->Please review and comment.

->PFB the screenshots of the output of the programs:-

**1**
Annotation 2020-06-30 155255
Annotation 2020-06-30 155212

**2**
Output_FibonacciN

### DIFF
--- a/src/nikhilMJun20/FibonacciN.java
+++ b/src/nikhilMJun20/FibonacciN.java
@@ -1,0 +1,42 @@
+/*To print first n Fibonacci Numbers.
+Input n = 5
+Output = 1 1 2 3 5 8
+
+Input: n = 10
+Output : 1 1 2 3 5 8 13 21 34 55
+
+Hint :
+The next number is found by adding up the two numbers before it:
+the 2 is found by adding the two numbers before it (1+1),
+the 3 is found by adding the two numbers before it (1+2),
+the 5 is (2+3),
+and so on!
+*/
+
+package nikhilMJun20;
+
+import java.util.Scanner;
+
+public class FibonacciN {
+	void displayFibonacci(int order) {
+		System.out.println("\nFibonacci Series of Order "+order+":-\n");
+		int num1=1,num2=1,num3=0;
+		for(int index=0; index<order; index++) {
+			if(index<2)
+				System.out.print(num1+" ");
+			else {
+				num3 = num1 + num2;
+				num1 = num2;
+				num2 = num3;
+				System.out.print((num3)+" ");
+			}
+		}
+	}
+	
+	public static void main(String[] args) {
+		Scanner scanner = new Scanner(System.in);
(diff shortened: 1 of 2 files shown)


src/nikhilMJun20/FibonacciN.java
```diff
@@ -0,0 +1,42 @@
...
+package nikhilMJun20;
+
+import java.util.Scanner;
+
+public class FibonacciN {
+	void displayFibonacci(int order) {
+		System.out.println("\nFibonacci Series of Order "+order+":-\n");
+		int num1=1,num2=1,num3=0;
+		for(int index=0; index<=order; index++) {
+			if(index<2)
```
VaishnaviVJun20, line 25: --You can assign 0 to num1 variable and then there is no need for this if else logic so you can reduce some lines of code.

src/nikhilMJun20/NumberSwap.java
```diff
@@ -0,0 +1,21 @@
...
+		System.out.println("Originally Given Numbers:-\nNumber1 -> "+num1+"     Number2 -> "+num2);
+		num1 += num2;
+		num2 = num1 - num2;
+		num1 -= num2;
+		System.out.println("\nSwapped Numbers:-\nNumber1 -> "+num1+"     Number2 -> "+num2);
+	}
+	
+	public static void main(String[] args) {
+		int num1=25, num2=35;
+		new NumberSwap().swapNum(num1, num2);
```
VaishnaviVJun20, line 19: --Code looks good to me, I like how you directly called method.

VaishnaviVJun20 (commented):
--Please work on comments.

src/nikhilMJun20/NumberSwap.java
NikhilMJun20: Thank you, Vaishnavi!

src/nikhilMJun20/FibonacciN.java
NikhilMJun20: ->Sure. I can make that change which is more efficient.
But this requires additional print statement outside the loop - which is kind of like an adjustment made by us without implementing logic of Fibonacci series. Logic should be like unless there are 2 elements which we can add, we should not proceed to 3rd element. 
Let me know your thoughts on this.

VaishnaviVJun20: -- Well if logically thinking this makes sense.
--However if we implement suggested logic, code can be little efficient.
-- I am also not sure which approach is better, so let's our reviewer take the call.

VaishnaviVJun20 (approved):
--Code looks good to me.
--Keep it up

src/nikhilMJun20/FibonacciN.java
KrishnaKTechnocredits: void printFibonacciNumber(int number) {
		int num =0; 
		int num1 = 0, num2=1;
		for(int index=1; index<=5; index++) {
			System.out.print(num2+" ");
			num = num1 + num2;
			num1= num2;
		    num2 = num;
		}
	}
Nikhil if block is not required.

PR #608, opened 2020-06-30
Piyush_Assignment_21_30Min

PiyushPJun20:
image

image

### DIFF
--- a/src/Fibonacci.java
+++ b/src/Fibonacci.java
@@ -3,7 +3,7 @@ import java.util.Scanner;
 public class Fibonacci {
 
 	public static void main(String[] args) {
-		int n, first = 0, next = 1;
+		int n, first = 1, next = 1;
 		System.out.println(" Enter the value of n : ");
 		Scanner scanner = new Scanner(System.in);
 		n = scanner.nextInt();
--- a/src/Fibonacci.java
+++ b/src/Fibonacci.java
@@ -1,0 +1,21 @@
+import java.util.Scanner;
+
+public class Fibonacci {
+
+	public static void main(String[] args) {
+		int n, first = 0, next = 1;
+		System.out.println(" Enter the value of n : ");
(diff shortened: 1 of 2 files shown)


src/Fibonacci.java
```diff
@@ -0,0 +1,21 @@
...
+public class Fibonacci {
+
+	public static void main(String[] args) {
+		int n, first = 0, next = 1;
+		System.out.println(" Enter the value of n : ");
+		Scanner scanner = new Scanner(System.in);
+		n = scanner.nextInt();
+		System.out.print(" Fibonacci Series is : ");
+		System.out.print(first + " " + next);
+		for (int i = 1; i <= n - 2; i++) {
```
ShrutiDJun20, line 12: As per requirement, our output should start from 1 and end on 55 when user enters 10
your logic is correct, just not matching with the expected output.

src/Swap.java
```diff
@@ -0,0 +1,14 @@
+
+public class Swap {
+	static void swapNum(int num1, int num2) {
+		num2 = num1 + num2;
+		num1 = num2 - num1;
+		num2 = num2 - num1;
+		System.out.println(" Numbers After Swapping : " + (num1) + "," + num2);
```
ShrutiDJun20: No need to put num1 in bracket here I think

src/Fibonacci.java
PiyushPJun20: Changes Incorporated.

src/Swap.java
PiyushPJun20: Thanks for suggestion, Changes Incorporated.

src/Fibonacci.java
PiyushPJun20: image

ShrutiDJun20 (approved):

PR #610, opened 2020-06-30
Vaishnavi_Assignment_21_25Min_Done

VaishnaviVJun20:
image
image

### DIFF
--- a/src/vaishnaviVJun20_arrayExamples/Fibonacci.java
+++ b/src/vaishnaviVJun20_arrayExamples/Fibonacci.java
@@ -18,7 +18,6 @@ public class Fibonacci {
 	
 	public static void main(String[] args) {
 		Fibonacci fibonacci = new Fibonacci();
-        int input=10;
-        fibonacci.printFibonacciNumbers(input);
+        fibonacci.printFibonacciNumbers(10);
 	}
 }
--- a/src/vaishnaviVJun20_arrayExamples/Fibonacci.java
+++ b/src/vaishnaviVJun20_arrayExamples/Fibonacci.java
@@ -1,0 +1,24 @@
+package vaishnaviVJun20_arrayExamples;
+
+//To print first n Fibonacci Numbers
+
+public class Fibonacci {
+	
+	void printFibonacciNumbers(int input) {
+		int sum=0,num1=0,num2=1;
+		System.out.println("Input number is:" +input);
+		System.out.print(num2);
+		for(int index=1;index<=input;index++) {
+			sum = num1+num2;
+			System.out.print(" " +sum +" ");
+			num1=num2;
+			num2=sum;			
+		}
+	}
+	
+	public static void main(String[] args) {
+		Fibonacci fibonacci = new Fibonacci();
+        int input=10;
+        fibonacci.printFibonacciNumbers(input);
+	}
+}
--- a/src/vaishnaviVJun20_arrayExamples/Fibonacci.java
+++ b/src/vaishnaviVJun20_arrayExamples/Fibonacci.java
@@ -7,10 +7,9 @@ public class Fibonacci {
 	void printFibonacciNumbers(int input) {
 		int sum=0,num1=0,num2=1;
 		System.out.println("Input number is:" +input);
-		System.out.print(num2);
 		for(int index=1;index<=input;index++) {
+			System.out.print(num2 +" ");
 			sum = num1+num2;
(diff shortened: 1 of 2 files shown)


src/vaishnaviVJun20_arrayExamples/Fibonacci.java
```diff
@@ -0,0 +1,24 @@
...
+			sum = num1+num2;
+			System.out.print(" " +sum +" ");
+			num1=num2;
+			num2=sum;			
+		}
+	}
+	
+	public static void main(String[] args) {
+		Fibonacci fibonacci = new Fibonacci();
+        int input=10;
```
AditiGJUN20: no need to initialize variable. You can directly pass the value in next step during method calling.
**fibonacci.printFibonacciNumbers(10)**

src/vaishnaviVJun20_arrayExamples/Fibonacci.java
```diff
@@ -0,0 +1,24 @@
+package vaishnaviVJun20_arrayExamples;
+
+//To print first n Fibonacci Numbers
+
+public class Fibonacci {
+	
+	void printFibonacciNumbers(int input) {
```
AditiGJUN20, line 7: The current logic is giving **11** elements instead of **10**. Check output screenshot.
**1 1  2  3  5  8  13  21  34  55  89**

src/vaishnaviVJun20_arrayExamples/Fibonacci.java
```diff
@@ -0,0 +1,24 @@
...
+		int sum=0,num1=0,num2=1;
+		System.out.println("Input number is:" +input);
+		System.out.print(num2);
```
AditiGJUN20: If printing sum no need to print num2.

src/vaishnaviVJun20_arrayExamples/Fibonacci.java
```diff
@@ -0,0 +1,24 @@
...
+		for(int index=1;index<=input;index++) {
+			sum = num1+num2;
+			System.out.print(" " +sum +" ");
+			num1=num2;
```
AditiGJUN20, line 13: make a little change in line 15,16 assignments and you'll get the output.

src/vaishnaviVJun20_arrayExamples/SwapTwoNum.java
```diff
@@ -0,0 +1,20 @@
...
+		System.out.println("Values before swapping: X=" +num1 +", Y="+num2);
+		num1 = num1+num2;
+		num2 = num1-num2;
+		num1 = num1-num2;
+		System.out.println("Values after swapping: X=" +num1 +", Y="+num2);
+	}
+
+	public static void main(String[] args) {
+		SwapTwoNum swapTwoNum = new SwapTwoNum();
+		int num1 = 10, num2=20;
```
AditiGJUN20: no need to initialize variable. You can directly pass the value in next step during method calling.
**swapTwoNum.swapTwoNumbers(10, 20)**

AditiGJUN20 (changes requested):
**For fibonacci series program it's returning 11 elements instead of 10.**

src/vaishnaviVJun20_arrayExamples/Fibonacci.java
VaishnaviVJun20: -- Thanks for the suggestion, I have made the changes.

src/vaishnaviVJun20_arrayExamples/Fibonacci.java
VaishnaviVJun20: -- In requirement it is mentioned that we need to print first and fibonnaci numbers.
--So according to given output as well I think we do need to print 11 numbers.

src/vaishnaviVJun20_arrayExamples/Fibonacci.java
VaishnaviVJun20: -- Printing num2 for required 11 numbers output.
--Let me know your thoughts on this

src/vaishnaviVJun20_arrayExamples/SwapTwoNum.java
VaishnaviVJun20: -- Changes done

AditiGJUN20 (approved):
Code Approved

PR #623, opened 2020-06-30
Viresh_Assignment_21_Done_(20+15)Mins

VireshJJun20:
Please review and comments.
image

image

### DIFF
--- a/src/vireshJJun20/Assignment21/FindFibonacci.java
+++ b/src/vireshJJun20/Assignment21/FindFibonacci.java
@@ -1,0 +1,36 @@
+/*
+Program 2: To print first n Fibonacci Numbers.
+Input n = 5
+Output = 1 1 2  3  5  8 
+
+Input: n = 10 
+Output :  1 1 2  3  5  8 13  21 34 55
+
+Hint :
+The next number is found by adding up the two numbers before it:
+the 2 is found by adding the two numbers before it (1+1),
+the 3 is found by adding the two numbers before it (1+2),
+the 5 is (2+3),
+and so on! 
+*/
+
+package vireshJJun20.Assignment21;
+
+public class FindFibonacci {
+
+	void getFibonacci(int input) {
+		int cnt1 = 0;
+		int cnt2 = 1;
+		int output = 0;
+		for (int i = 1; i <= input; i++) {
+			output = cnt1 + cnt2;
+			cnt1 = cnt2;
+			cnt2 = output;
+			System.out.print(cnt1 + " ");
+		}
+	}
+
+	public static void main(String[] args) {
+		new FindFibonacci().getFibonacci(10);
+	}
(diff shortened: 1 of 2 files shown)


src/vireshJJun20/Assignment21/FindFibonacci.java
```diff
@@ -0,0 +1,36 @@
...
+and so on! 
+*/
+
+package vireshJJun20.Assignment21;
+
+public class FindFibonacci {
+
+	void getFibonacci(int input) {
+		int cnt1 = 0;
+		int cnt2 = 1;
```
DeavinaSJun20, line 23: Can we change variables names to something like firstNumber & secondNumber

src/vireshJJun20/Assignment21/FindFibonacci.java
```diff
@@ -0,0 +1,36 @@
...
+		for (int i = 1; i <= input; i++) {
+			output = cnt1 + cnt2;
+			cnt1 = cnt2;
+			cnt2 = output;
+			System.out.print(cnt1 + " ");
+		}
+	}
+
+	public static void main(String[] args) {
+		new FindFibonacci().getFibonacci(10);
```
DeavinaSJun20, line 34: Please take input from user and not hardcode the value that we need to print first 10 numbers only. Logic seems Ok

src/vireshJJun20/Assignment21/NumberSwap.java
```diff
@@ -0,0 +1,22 @@
...
+	void getSwappedNumber(int num1, int num2) {
+		System.out.println("Before Swapping: " + num1 + " " + num2);
+		num1 += num2;
+		num2 = num1 - num2;
+		num1 = num1 - num2; 
+		System.out.println("After Swapping: " + num1 + " " + num2);
+	}
+
+	public static void main(String[] args) {
+		new NumberSwap().getSwappedNumber(10, 20);
```
DeavinaSJun20, line 20: take values from user instead of just hard-coding the inputs. Logic looks good

src/vireshJJun20/Assignment21/FindFibonacci.java
VireshJJun20: Ok.

src/vireshJJun20/Assignment21/FindFibonacci.java
VireshJJun20: Do you mean using Scanner class?

DeavinaSJun20: yes

KrishnaKTechnocredits: Devina Here no need to use scanner class. I suggested hard code value only.

PR #626, opened 2020-06-30
Suparna_Assignment_21_10min

SuparnaNikam:
maxnoand febonacci

### DIFF
--- a/src/suparnaNJune20/Assignment_21/MaxNoAndFibonacci.java
+++ b/src/suparnaNJune20/Assignment_21/MaxNoAndFibonacci.java
@@ -1,0 +1,47 @@
+/*
+ * Assignment 21 : 30th Jun 2020
+
+Program 1: Swap two numbers without using third variable.
+Input: num1= 10, num2 =20
+output: num1=20, num2 =10
+
+Program 2: To print first n Fibonacci Numbers.
+Input n = 6
+Output = 1 1 2  3  5  8
+
+Input: n = 10 
+Output :  1 1 2  3  5  8 13  21 34 55
+
+ */
+package suparnaNJune20.Assignment_21;
+
+public class MaxNoAndFibonacci {
+
+	void maxNo(int no1, int no2) {
+		System.out.println("Before swap No1 = " + no1 + " No2 = " + no2);
+		no1 = no1 + no2;
+		no2 = no1 - no2;
+		no1 = no1 - no2;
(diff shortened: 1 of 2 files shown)


src/suparnaNJune20/Assignment_21/MaxNoAndFibonacci.java
```diff
@@ -0,0 +1,47 @@
...
+	void maxNo(int no1, int no2) {
+		System.out.println("Before swap No1 = " + no1 + " No2 = " + no2);
+		no1 = no1 + no2;
+		no2 = no1 - no2;
+		no1 = no1 - no2;
+		System.out.println("After swap No1 = " + no1 + " No2 = " + no2);
+	}
+
+	void fibonacciSeries(int count) {
+		int no1 = 0, no2 = 1, no3, index;
```
VireshJJun20: As per your current logic, Fibonacci series is starting from 0. 
Also, if value of n= 6, then ideally it should reflect 6 numbers. same when n=10 
you can try these changes:
int no1 = 1;

src/suparnaNJune20/Assignment_21/MaxNoAndFibonacci.java
```diff
@@ -0,0 +1,47 @@
...
+		System.out.print(no1 + " " + no2);
+		for (index = 2; index <= count; ++index) {
```
VireshJJun20: Check with this: 
for (index = 2; index < count; ++index) 

Rest logic looks good to me.

VireshJJun20: Check with this: 
for (index = 2; index < count; ++index)

src/suparnaNJune20/Assignment_21/MaxNoAndFibonacci.java
```diff
@@ -0,0 +1,47 @@
...
+Input n = 6
+Output = 1 1 2  3  5  8
+
+Input: n = 10 
+Output :  1 1 2  3  5  8 13  21 34 55
+
+ */
+package suparnaNJune20.Assignment_21;
+
+public class MaxNoAndFibonacci {
```
VireshJJun20, line 15: Requirement is to create 2 separate programs.

SuparnaNikam: created 2 separate program.

PR #632, opened 2020-06-30
Madhura_Assignment_21_1hr

MadhuraMJun20:
Find Fibonacci numbers : 
FibonacciNumbers
Swapping of two numbers : 
SwapTwoNumbers

### DIFF
--- a/src/madhuraMJun20/scannerClass/FibonacciNumbers.java
+++ b/src/madhuraMJun20/scannerClass/FibonacciNumbers.java
@@ -1,0 +1,29 @@
+package madhuraMJun20.scannerClass;
+
+/*Program 2: To print first n Fibonacci Numbers.
+Input n = 6
+Output = 1 1 2  3  5  8
+
+Input: n = 10 
+Output :  1 1 2  3  5  8 13  21 34 55*/
+public class FibonacciNumbers {
+	void findFibonacciNumber(int n) {
+		int firstNum = 0;
+		int nextNum = 1;
+		System.out.print("\nFirst " + n + " Fibonacci numbers :- \n");
+
+		for (int index = 0; index < n; index++) {
+			int sum = firstNum + nextNum;
+			System.out.print(nextNum + " ");
+			firstNum = nextNum;
+			nextNum = sum;
+		}
+	}
+
+	public static void main(String[] args) {
+		FibonacciNumbers fibonacciNumbers = new FibonacciNumbers();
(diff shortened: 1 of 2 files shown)


src/madhuraMJun20/scannerClass/FibonacciNumbers.java
```diff
@@ -0,0 +1,29 @@
...
+Input: n = 10 
+Output :  1 1 2  3  5  8 13  21 34 55*/
+public class FibonacciNumbers {
+	void findFibonacciNumber(int n) {
+		int firstNum = 0;
+		int nextNum = 1;
+		System.out.print("\nFirst " + n + " Fibonacci numbers :- \n");
+
+		for (int index = 0; index < n; index++) {
+			int sum = firstNum + nextNum;
```
HarshadBJun20, line 16: Code logic is good. I suggest you can reduce the code without using the extra variable "sum" here. Think about it.

src/madhuraMJun20/scannerClass/SwapTwoNumbers.java
```diff
@@ -0,0 +1,19 @@
...
+		System.out.println("Input -> Number 1 = " + num1 + ", Number 2 = " + num2);
+		num2 = num1 + num2;// 30
+		num1 = num2 - num1;// 20
+		num2 = num2 - num1;
+		System.out.println("Output -> Number 1 = " + num1 + ", Number 2 = " + num2);
+	}
+
+	public static void main(String[] args) {
+		SwapTwoNumbers swapTwoNumbers = new SwapTwoNumbers();
+		int num1 = 10;
```
HarshadBJun20: Code is perfect. Just one thing that I would suggest is you should pass the numbers directly to the method calling instead of declaring 2 variables and assigning the value.

HarshadBJun20 (approved):
I am Approving the PR as code looks good to me. But, I would suggest to try the requested logic for Fibonacci code.

src/madhuraMJun20/scannerClass/SwapTwoNumbers.java
MadhuraMJun20: Yes. Updated the code. thanks

src/madhuraMJun20/scannerClass/FibonacciNumbers.java
MadhuraMJun20: Ok. I will try

PR #636, opened 2020-07-01
Rachana_Assignment_22_Done_18min

RachanaGJun20:
--------- Check Number is Palindrome----------------
palindrome22
---------------------------------------------------------------------
--------- Check String is Palindrome----------------
stringplai22

### DIFF
--- a/src/rachanaGJun20/PalindromeNumber.java
+++ b/src/rachanaGJun20/PalindromeNumber.java
@@ -1,0 +1,53 @@
+/*Program 1: 
+Verify given number is palindrome or not. 
+Input: number = 45654
+Output : Given number 45654 is palindrome.
+
+Input: number = 56789
+Output : Given number 56789 is not palindrome.*/
+
+package rachanaGJun20;
+
+import java.util.Scanner;
+
+public class PalindromeNumber {
+	//method will check whether given number is Palindrome
+	boolean isNumberPalindrome(int number) {
+		int remainder = 0;
+		int reverseNumber = 0;
+		int givenNumber = number;
+
+		while (number != 0) {
+			remainder = number % 10;
+			number = number / 10;
+			reverseNumber = reverseNumber * 10 + remainder;
+		}
+		if (givenNumber == reverseNumber)
+			return true;
+		else
+			return false;
+	}
+
+	public static void main(String[] args) {
+		PalindromeNumber palindromenumber = new PalindromeNumber();
+		Scanner scanner = new Scanner(System.in);
+		boolean flag;
+		System.out.println("Please enter Number to check");
+		flag = palindromenumber.isNumberPalindrome(scanner.nextInt());
+		if (flag == true) {
+			System.out.println("Number is palindrome");
+		} else {
+			System.out.println("Number is not palindrome");
+		}
+		System.out.println("Please enter Number to check");
+		flag = palindromenumber.isNumberPalindrome(scanner.nextInt());
+		if (flag == true) {
+			System.out.println("Number is palindrome");
+		} else {
+			System.out.println("Number is not palindrome");
+		}
+		scanner.close();
+
+	}
+
+}
--- a/src/rachanaGJun20/PalindromeNumber.java
+++ b/src/rachanaGJun20/PalindromeNumber.java
@@ -34,14 +34,14 @@ public class PalindromeNumber {
 		boolean flag;
 		System.out.println("Please enter Number to check");
 		flag = palindromenumber.isNumberPalindrome(scanner.nextInt());
-		if (flag == true) {
+		if (flag) {
 			System.out.println("Number is palindrome");
 		} else {
 			System.out.println("Number is not palindrome");
 		}
 		System.out.println("Please enter Number to check");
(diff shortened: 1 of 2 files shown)


src/rachanaGJun20/PlindromeString.java
```diff
@@ -0,0 +1,41 @@
+package rachanaGJun20;
+
+import java.util.Scanner;
+
+public class PlindromeString {
+	//method will check whether given string is Palindrome
+	boolean isStringPlindrome(String str) {
+		String revstr = "";
+		String givenString = str;
```
AjitSJune20: Suggestion 1: Instead of assigning value to new string , we can use directly Str value(In line 14) as well

src/rachanaGJun20/PlindromeString.java
```diff
@@ -0,0 +1,41 @@
...
+
+		for (int i = str.length() - 1; i >= 0; i--) {
+			revstr = revstr + str.charAt(i);
+		}
+		if (givenString.equals(revstr))
```
AjitSJune20: Suggestion 2: use equalsIgnoreCase

src/rachanaGJun20/PlindromeString.java
```diff
@@ -0,0 +1,41 @@
...
+			return true;
+		else
+			return false;
+	}
+	
+	public static void main(String[] args) {
+		PlindromeString plindromestring = new PlindromeString();
+		Scanner scanner = new Scanner(System.in);
+		boolean flag;
+		System.out.println("Pease Enter a string to check:");
```
AjitSJune20, line 22: Suggestion 3: to reduce line of code (Line no 22 to 38)
you can write like this as well to reduce 15 lines of code into  6 lines

Rough Code : 
String[] stringArray = { "radar", "technocredits" };
		for (int i=0;i<stringArray.length;i++) {
		String string=stringArray[i];
			if (isStringPlindrome(string) == true)   //here method calling statement  depends on 
                                                                                     //Static or no Static
				System.out.println("Given String " + string + " is palindrome");
			else
				System.out.println("Given String " + string + " is not palindrome");
		}

src/rachanaGJun20/PalindromeNumber.java
```diff
@@ -0,0 +1,53 @@
...
+
+	public static void main(String[] args) {
+		PalindromeNumber palindromenumber = new PalindromeNumber();
+		Scanner scanner = new Scanner(System.in);
+		boolean flag;
+		System.out.println("Please enter Number to check");
+		flag = palindromenumber.isNumberPalindrome(scanner.nextInt());
+		if (flag == true) {
+			System.out.println("Number is palindrome");
+		} else {
```
AjitSJune20, line 39: Suggestion same as PalindromeString program.

src/rachanaGJun20/PlindromeString.java
RachanaGJun20: Thanks Ajit , Yes got it will make that change

src/rachanaGJun20/PlindromeString.java
RachanaGJun20: I think , In reversing a string and checking for palindrome we should have to use string as it is in the given format.
we can not use equalsIgnoreCase here I think.

src/rachanaGJun20/PalindromeNumber.java
RachanaGJun20: Do you want to suggest about givennumber variable ? if no plz ignore below lines and if yes then
Here in my isNumberPalindrom logic I need to assign number into givennumber variable otherwise I won't able to match it. beacuse  number variable is getting change while performing operations in while loop.

and same I want to say about taking user input as commented on Palindrom string.

src/rachanaGJun20/PlindromeString.java
RachanaGJun20: one thing I can surely do like I am repeating scanning two times,for this I can write loop,so,few lines for code I can reduce.

RachanaGJun20: Thanks for your input Ajit !
I will make these chnages as suggested.if (isStringPlindrome(string) == true)
But as  always try to make generalized logic for every possible program so that in future that program will get run for any given string without making any change in code. so not defining hard coded array .

KrishnaKTechnocredits: @RachanaGJun20 you can hardcode values, your method logic should be independent of hardcode values. from main method if you pass hardcode value that is accepted. Now you comfortable with scanner class, so now no need to use it.

KrishnaKTechnocredits (approved):
Code looks good to me but no more use of scanner class. It will save your time. Lets invest time in building logic, ideally in selenium we never use scanner class. Scanner class is only use for interview or when you create cmd line utility.

RachanaGJun20:
> Code looks good to me but no more use of scanner class. It will save your time. Lets invest time in building logic, ideally in selenium we never use scanner class. Scanner class is only use for interview or when you create cmd line utility.

Sure krishna......  :)

PR #638, opened 2020-07-01
Nikhil_Assignment_22_Done_(5+15=)20Min

NikhilMJun20:
->Completed and pushed Assignment # 22.
->Please review and comment.

->PFB screenshots of the outputs of the programs:-

**1**
Output_PalindromeNumber

**2**
Output_PalindromeString

### DIFF
--- a/src/nikhilMJun20/PalindromeNumber.java
+++ b/src/nikhilMJun20/PalindromeNumber.java
@@ -1,0 +1,39 @@
+/*Verify given number is palindrome or not. 
+Input: number = 45654
+Output : Given number 45654 is palindrome.
+
+Input: number = 56789
+Output : Given number 56789 is not palindrome.
+*/
+
+package nikhilMJun20;
+
+import java.util.Scanner;
+
+public class PalindromeNumber {
+	void checkPalindromeNumber(int number) {
+		int revNumber = 0, originalNumber=number;
+		while(number>0) {
+			revNumber = 10*revNumber+(number%10);
+			number = number/10;
+		}
+		if(revNumber == originalNumber)
+			System.out.println(revNumber+" is a Palindrome number.");
+		else
+			System.out.println(revNumber+" is NOT a Palindrome number.");
+	}
+	
+	public static void main(String[] args) {
+		PalindromeNumber palindromeNumber = new PalindromeNumber();
+		Scanner scanner = new Scanner(System.in);
+		String proceed = "y";
+		System.out.println("This program will check if entered number is Palindrome.\n");
+		do {
+			System.out.println("Enter a number:");
+			palindromeNumber.checkPalindromeNumber(scanner.nextInt());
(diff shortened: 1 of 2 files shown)


src/nikhilMJun20/PalindromeString.java
```diff
@@ -0,0 +1,42 @@
...
+
+package nikhilMJun20;
+
+import java.util.Scanner;
+
+public class PalindromeString {
+	void checkPalindromeString(String str) {
+		boolean flag = true;
+		str = str.toLowerCase();
+		for(int leftIndex=0, rightIndex=str.length()-1; leftIndex<=rightIndex; leftIndex++,rightIndex--) {
```
VaishnaviVJun20, line 17: -- Your code and logic looks good to me.
--However I think you can simplify it a little.
--Refer below code:
String reverse="", originalName = name;
		for(int index=name.length()-1;index>=0;index--) {
			reverse = reverse+name.charAt(index);		
		}
		if(originalName.equals(reverse))
			System.out.println(" Given String " +originalName +" is Palindrome");
		else
			System.out.println(" Given String " +originalName +" is not Palindrome");

VaishnaviVJun20 (approved):
--Code looks good to me

src/nikhilMJun20/PalindromeString.java
NikhilMJun20: ->Yes, it is a simple and string reverse logic. But, in my logic I do not required an additional string for keeping the reversed string and at the first mismatch - where one incrementing and one decrementing index pointer is used, I will be able to tell that the string is not a palindrome.
->Thanks for the review comment.

KrishnaKTechnocredits: @VaishnaviVJun20 Nikhil's logic is correct and as expected. Not required another string to store all characters in reverse order.

src/nikhilMJun20/PalindromeNumber.java
```diff
@@ -0,0 +1,39 @@
...
+package nikhilMJun20;
+
+import java.util.Scanner;
+
+public class PalindromeNumber {
+	void checkPalindromeNumber(int number) {
+		int revNumber = 0, originalNumber=number;
+		while(number>0) {
+			revNumber = 10*revNumber+(number%10);
+			number = number/10;
```
KrishnaKTechnocredits, line 18: To reduce this line, you can write operation on while condition like below
while((number = number/10)>0)

KrishnaKTechnocredits (approved):
Code looks good to me.

src/nikhilMJun20/PalindromeNumber.java
NikhilMJun20: Yes. This becomes even more compact.
Next time on I will use this style whenever possible.
Thank you, @KrishnaKTechnocredits !

PR #640, opened 2020-07-01
ADT_Assignment_22_Done_15_10Mins

AditiGJUN20:
**Assignment 22**
**1. PalindromeNumber.java**
Time: 10Mins
Screen Shot 2020-07-01 at 1 10 00 AM

**2. PalindromeString.java **
Time: 15Mins
Screen Shot 2020-07-01 at 1 23 21 AM


AmitaRJun20 (approved):
Codes are perfectly written as per expectation... Good job...

src/aditiGJUN20/PalindromeNumber.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class PalindromeNumber {
+
+	// Method to check number is palindrome or not.
+	void checkPalindromeNumber(int num) {
+		int reverseNum = 0, inputNum = num;
+		System.out.println("\nInput Number = " + inputNum);
+		while (num > 0) {
+			reverseNum = reverseNum * 10 + (num % 10);
+			num = num / 10;
```
KrishnaKTechnocredits, line 22: we can write assignment operation in while loop. Just FYI.

while((number = number/10)>0)

src/aditiGJUN20/PalindromeString.java
```diff
@@ -0,0 +1,30 @@
...
+package aditiGJUN20;
+
+public class PalindromeString {
+	boolean isPalindromString(String input) {
+		System.out.println("Input String: "+input);
+		String rev="";
+		for ( int index = input.length() - 1; index >= 0; index-- )
+        rev = rev + input.charAt(index); 
+		
+		if(input.equalsIgnoreCase(rev))
```
KrishnaKTechnocredits, line 17: Better to use equal method. If interviewer asked to checking it without case sensitivity you should equalsIgnoreCase

KrishnaKTechnocredits (approved):
Code looks good to me. Please take a look at comments.

src/aditiGJUN20/PalindromeString.java
```diff
@@ -0,0 +1,30 @@
...
+ * Output: Given string technocredits is not palindrome.
+*/
+package aditiGJUN20;
+
+public class PalindromeString {
+	boolean isPalindromString(String input) {
+		System.out.println("Input String: "+input);
+		String rev="";
+		for ( int index = input.length() - 1; index >= 0; index-- )
+        rev = rev + input.charAt(index);
```
KrishnaKTechnocredits, line 15: you can compare string from both end without having rev [extra] variable.

KrishnaKTechnocredits (approved):

src/aditiGJUN20/PalindromeNumber.java
AditiGJUN20: This is new.  i was aware about for loop. But not knowing that for while loop also we can create similar behavior.
Thank you Krishna.

src/aditiGJUN20/PalindromeString.java
AditiGJUN20: I'll keep that in mind.

PR #644, opened 2020-07-01
“Ajit_Assignment_22_(20+10 mins)”

AjitSJune20:
image
image

### DIFF
--- a/src/ajitSJune20/TestPalindromeNumber.java
+++ b/src/ajitSJune20/TestPalindromeNumber.java
@@ -1,0 +1,33 @@
+package ajitSJune20;
+
+public class TestPalindromeNumber {
+
+	static boolean isPalindromeInteger(int number) {
+		if (number == reverse(number)) {
+			return true;
+		}
+		return false;
+	}
+
+	static int reverse(int number) {
+		int reverse = 0;
+		while (number != 0) {
+			reverse = reverse * 10 + number % 10;
+			number = number / 10;
+		}
+		return reverse;
+	}
+
+	public static void main(String args[]) {
+
+		int[] intArray = { 45654, 56789 };
+		// for (int i=0;i<intArray.length;i++) {
+		// int number=intArray[i];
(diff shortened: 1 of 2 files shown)


src/ajitSJune20/TestPalindromeNumber.java
```diff
@@ -0,0 +1,33 @@
+package ajitSJune20;
+
+public class TestPalindromeNumber {
+
+	static boolean isPalindromeInteger(int number) {
```
KartikeyDJune20, line 5: in my opinion .. u are checking reverse number with provided number two times .. we can set both the things reverse+palindrome in main also..

AjitSJune20: yes I could have written every thing in main but for standard practice and guideline of Kirishna, we prefer to create method to perform operation/action. 
For re-usability of my code , I have written two separate method to check reverse and palindrome.  we can use same program to find reverse or palindrome. In our earlier assignment we wrote a program to find reverse, I enhanced same program to find palindrome.

KartikeyDJune20: Okay np.. bro.. its good u used the same function for both..

KrishnaKTechnocredits: Code Reusability looks good to me @AjitSJune20

src/ajitSJune20/TestPalindromeString.java
```diff
@@ -0,0 +1,30 @@
+package ajitSJune20;
+
+public class TestPalindromeString {
+	static String revStr = "";
+
+	static boolean isPalindromeString(String string) {
+		if (string.equalsIgnoreCase(reverse(string))) {
```
KrishnaKTechnocredits, line 7: better to use equals method.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #646, opened 2020-07-01
Aavruti_Assignment_22_(5+5)10mins

AavrutiDJun20:
Program 1 : Verify given number is Palindrome or not

Program 2 : Verify given string is Palindrome or not 

PalindromeNumber

PalindromeString

### DIFF
--- a/src/aavrutiDJun20/PalindromeString.java
+++ b/src/aavrutiDJun20/PalindromeString.java
@@ -1,0 +1,25 @@
+//Print whether given String is Palindrome or not
+package aavrutiDJun20;
+
+public class PalindromeString {
+	
+	void palindromeStringOrNot(String input) {
+		String palindromeInput = "";
+		
+		for(int index=input.length()-1;index>=0;index--) {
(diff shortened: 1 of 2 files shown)


YashCJun20 (approved):
Code Looks Perfect

src/aavrutiDJun20/PalindromeString.java
```diff
@@ -0,0 +1,25 @@
...
+public class PalindromeString {
+	
+	void palindromeStringOrNot(String input) {
+		String palindromeInput = "";
+		
+		for(int index=input.length()-1;index>=0;index--) {
+			palindromeInput += String.valueOf(input.charAt(index));
+		}
+		
+		if(palindromeInput.equalsIgnoreCase(input))
```
KrishnaKTechnocredits, line 13: I believe we should use equals(), if interviewer ask to make it case-insensitive we can use equalIgnoreCase

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #649, opened 2020-07-01
Kartikey_Assignment_22_Done_12mins

KartikeyDJune20:
image
image

### DIFF
--- a/src/kartikeyDJune20/PalindromeString.java
+++ b/src/kartikeyDJune20/PalindromeString.java
@@ -1,0 +1,21 @@
+package kartikeyDJune20;
+
+public class PalindromeString {
+
+	static void checkStringPalindrome(String input) {
+		String rev = "";
+		for (int index = input.length() - 1; index >= 0; index--) {
+			rev = rev + String.valueOf(input.charAt(index));
+		}
+		if (rev.equalsIgnoreCase(input))
+			System.out.println("Given string " + input + " is Palindrome");
+		else
+			System.out.println("Given string " + input + " is not Palindrome");
+	}
+
+	public static void main(String[] args) {
+		PalindromeString.checkStringPalindrome("technocredits");
+		PalindromeString.checkStringPalindrome("Radar");
+	}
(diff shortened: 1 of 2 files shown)


src/kartikeyDJune20/PalindromeString.java
```diff
@@ -0,0 +1,21 @@
+package kartikeyDJune20;
+
+public class PalindromeString {
+
+	static void checkStringPalindrome(String input) {
+		String rev = "";
+		for (int index = input.length() - 1; index >= 0; index--) {
+			rev = rev + String.valueOf(input.charAt(index));
```
harshvegada: Here, we don't need of String.valurOf() method as we concat the string first and then char, but if it's reverse then we will go this method. No impact on logic .

KartikeyDJune20: yes yes.. Thanku Harsh.. it should be optimized.

KartikeyDJune20: Changes Incorporated..

harshvegada (commented):
One correction @KartikeyDJune20 
Here we are checking exact to exact string if both case match then only pass the condition 
Do not use equvalsIgnore case 
Correct the changes

src/kartikeyDJune20/PalindromeString.java
```diff
@@ -0,0 +1,21 @@
...
+		}
+		if (rev.equals(input))
+			System.out.println("Given string " + input + " is Palindrome");
+		else
+			System.out.println("Given string " + input + " is not Palindrome");
+	}
+
+	public static void main(String[] args) {
+		PalindromeString.checkStringPalindrome("technocredits");
+		PalindromeString.checkStringPalindrome("Radar");
```
KrishnaKTechnocredits, line 18: R should be small otherwise it will say it is not palindrome :)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #653, opened 2020-07-01
AMITA_Assignment_22_Done_15mins

AmitaRJun20:
AMITA_Assignment_22 completed...please review...
PalindromeNumber
palindromeString

### DIFF
--- a/src/amitaRJun20/PalindromeNumber.java
+++ b/src/amitaRJun20/PalindromeNumber.java
@@ -1,0 +1,25 @@
+package amitaRJun20;
+
+public class PalindromeNumber {
+
+	void isPalindromeNumber(int num) {
+		System.out.println("The Input number is " + num);
+		int reverseNum = 0, inputnum = num;
+		while (num > 0) {
+			reverseNum = (reverseNum * 10) + (num % 10);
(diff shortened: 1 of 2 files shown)


src/amitaRJun20/PalindromeNumber.java
```diff
@@ -0,0 +1,25 @@
+package amitaRJun20;
+
+public class PalindromeNumber {
+
+	void isPalindromeNumber(int num) {
```
MaheshKJun20, line 5: Suggestion: Can add a comment to describe the method. It will be helpful when we are trying to find code or some else is trying to understand your code base which is huge.
Rest all is perfect.

MaheshKJun20 (approved):
Hi Amita,
Reviewed your Code it is as per the expectation. 
Approving it for further review.

src/amitaRJun20/PalindromeNumber.java
AmitaRJun20: Sure Mahesh...Thanks for your feedback...i will keep in mind from next time onwards...

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #655, opened 2020-07-01
Archana_Assignment_22_Done_10min+10min

ArchanaNJun20:
Palindrome Number 10min
image
Palindrome String10min
image

### DIFF
--- a/src/archanaNJun20/PalindromeNumber.java
+++ b/src/archanaNJun20/PalindromeNumber.java
@@ -1,9 +1,10 @@
 package archanaNJun20;
 
 public class PalindromeNumber {
-	int rev=0;
-	void palindromeNumber(int number){
+
+	void printPalindromeNumber(int number){
 		int oNumber=number;
+		int rev=0;
 		while(number>0){
 			int mod=number%10;
 			rev=(rev*10)+mod;
@@ -16,7 +17,7 @@ public class PalindromeNumber {
 	}
 	public static void main(String[] args) {
 		PalindromeNumber palindromeNumber=new PalindromeNumber();
-		palindromeNumber.palindromeNumber(45654);
-		palindromeNumber.palindromeNumber(56789);
+		palindromeNumber.printPalindromeNumber(45654);
+		palindromeNumber.printPalindromeNumber(56789);
 	}
 }
--- a/src/archanaNJun20/PalindromeNumber.java
+++ b/src/archanaNJun20/PalindromeNumber.java
@@ -1,0 +1,22 @@
+package archanaNJun20;
+
+public class PalindromeNumber {
+	int rev=0;
+	void palindromeNumber(int number){
+		int oNumber=number;
+		while(number>0){
+			int mod=number%10;
+			rev=(rev*10)+mod;
+			number=number/10;
+		}	
+		if(oNumber==rev)
+			System.out.println("Given number : "+ oNumber +" is palindrome");
+		else
+			System.out.println("Given number : "+ oNumber +" is not palindrome");
+	}
+	public static void main(String[] args) {
+		PalindromeNumber palindromeNumber=new PalindromeNumber();
+		palindromeNumber.palindromeNumber(45654);
+		palindromeNumber.palindromeNumber(56789);
+	}
+}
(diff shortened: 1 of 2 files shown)


src/archanaNJun20/PalindromeNumber.java
```diff
@@ -0,0 +1,22 @@
...
+		}	
+		if(oNumber==rev)
+			System.out.println("Given number : "+ oNumber +" is palindrome");
+		else
+			System.out.println("Given number : "+ oNumber +" is not palindrome");
+	}
+	public static void main(String[] args) {
+		PalindromeNumber palindromeNumber=new PalindromeNumber();
+		palindromeNumber.palindromeNumber(45654);
+		palindromeNumber.palindromeNumber(56789);
```
AavrutiDJun20: It is not best practice to reuse same variable name with different context.
Always try to use different name for method and variable name to maintain understandability and readability of your program.

src/archanaNJun20/PalindromeString.java
```diff
@@ -0,0 +1,23 @@
+package archanaNJun20;
+
+public class PalindromeString {
+	void palindromeString(String inputString){ 
+		String originalstring=inputString;
+		int length=inputString.length();
+		String newString = "";
+		for(int index=length-1;index>=0;index--){
+			char ch=inputString.charAt(index);
+			newString=newString+String.valueOf(ch);
```
AavrutiDJun20: you can also write like this newString += String.valueOf(inputString.charAt(index));

src/archanaNJun20/PalindromeString.java
```diff
@@ -0,0 +1,23 @@
...
+		}
+		if(newString.equalsIgnoreCase(originalstring))
```
AavrutiDJun20: you can directly use like this newString.equalsIgnoreCase(inputString)
instead of creating new variable "originalstring". As you are not changing value of inputString variable.

src/archanaNJun20/PalindromeNumber.java
```diff
@@ -0,0 +1,22 @@
+package archanaNJun20;
+
+public class PalindromeNumber {
+	int rev=0;
```
AavrutiDJun20: Instead of defining "rev" as instance variable you can define it in method itself as a local variable.

ArchanaNJun20: The code is updated based on the comment. Thank you for your Suggestion.

src/archanaNJun20/PalindromeString.java
ArchanaNJun20: The code is updated based on the comment. Thank you for your Suggestion.

src/archanaNJun20/PalindromeString.java
ArchanaNJun20: The code is updated based on the comment. Thank you for your Suggestion.

src/archanaNJun20/PalindromeNumber.java
ArchanaNJun20: The code is updated based on the comment. Thank you for your Suggestion.

AavrutiDJun20 (approved):
Good Job

src/archanaNJun20/PalindromeString.java
KrishnaKTechnocredits: equalsIgnoreCase never used in palindrome string. Always use .equals()

src/archanaNJun20/PalindromeNumber.java
```diff
@@ -0,0 +1,23 @@
+package archanaNJun20;
+
+public class PalindromeNumber {
+
+	void printPalindromeNumber(int number){
+		int oNumber=number;
+		int rev=0;
```
KrishnaKTechnocredits, line 7: What is rev ??? Is it understandbale name. Write like revNumber

src/archanaNJun20/PalindromeNumber.java
```diff
@@ -0,0 +1,23 @@
...
+		while(number>0){
+			int mod=number%10;
```
KrishnaKTechnocredits, line 9: Never initialize variable in loop. and also no need of this variable int mod. 
revNumber = (revNumber*10) + (number%10);
number=number/10;

PR #656, opened 2020-07-01
Mahesh_Assignment_22_15Mins_Done

MaheshKJun20:
PalindromeNumber
PalindromeString
Mahesh_Assignment_22_15Mins_Done


harshvegada (approved):
In PalindromeString program line 21, We should go for equals() as we want to check exact String,
Logic wise both are same, no comments on Logic part, you are pretty good in logic.
Keep it up! Just heads up for this little part

MaheshKJun20:
Thanks Harsh. Will keep in mind and implement in future assigments as well.

PR #664, opened 2020-07-01
Swapna_Assignment_22_Done_25_min

SwapnaWJun20:
image
image

### DIFF
--- a/src/swapnaWJun20/PalindromeNo.java
+++ b/src/swapnaWJun20/PalindromeNo.java
@@ -5,16 +5,14 @@ public class PalindromeNo {
 	void findPalindromeNumber(int number) {
 		int number1 = number;
 		int temp = 0;
-		int mod = 0;
 		while (number1 > 0) {
-			mod = number1 % 10;
+			temp = temp * 10 + number1 % 10;
 			number1 = number1 / 10;
-			temp = mod + temp * 10;
 		}
 		if (temp == number)
-			System.out.println("Input : "+number + " is Palindrome");
+			System.out.println("Input : " + number + " is Palindrome");
 		else
-			System.out.println("Input : "+number + " is not Palindrome");
+			System.out.println("Input : " + number + " is not Palindrome");
 	}
 
 	public static void main(String[] args) {
--- a/src/swapnaWJun20/PalindromeNo.java
+++ b/src/swapnaWJun20/PalindromeNo.java
@@ -1,0 +1,25 @@
+package swapnaWJun20;
+
+public class PalindromeNo {
+
+	void findPalindromeNumber(int number) {
+		int number1 = number;
+		int temp = 0;
+		int mod = 0;
+		while (number1 > 0) {
+			mod = number1 % 10;
+			number1 = number1 / 10;
+			temp = mod + temp * 10;
+		}
+		if (temp == number)
+			System.out.println("Input : "+number + " is Palindrome");
+		else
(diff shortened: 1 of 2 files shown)


src/swapnaWJun20/PalindromeNo.java
```diff
@@ -0,0 +1,25 @@
...
+
+	void findPalindromeNumber(int number) {
+		int number1 = number;
+		int temp = 0;
+		int mod = 0;
+		while (number1 > 0) {
+			mod = number1 % 10;
+			number1 = number1 / 10;
+			temp = mod + temp * 10;
+		}
```
AnupSJun20, line 11: Logic is correct but the code can be optimized and the mod variable is not required. instead, you can write in the below way. Please remove the commented line below and  update the logic you will get the result.

//int mod = 0;
		while (number1 > 0) {
			temp = temp * 10 + number1 % 10;
			number1 = number1 / 10;
			//temp = mod + temp * 10;
		}

src/swapnaWJun20/PalindromeString.java
```diff
@@ -0,0 +1,24 @@
+package swapnaWJun20;
+
+public class PalindromeString {
+
+	void findPalindromeNumber(String input) {
+		String input1 = input;
```
AnupSJun20: Here you don't need an extra variable to store the given string. the other variable reverse will work.

src/swapnaWJun20/PalindromeString.java
```diff
@@ -0,0 +1,24 @@
...
+		String reverse = "" ;
+
+		for (int index = input.length() - 1; index >= 0; index--) {
+			reverse += input.charAt(index);
+		}
+
+		if (input1.equals(reverse))
```
AnupSJun20: you can change input1 to input

AnupSJun20 (changes requested):
code looks good. Please make necessary changes. 
Remove extra lines where ever it is there.

src/swapnaWJun20/PalindromeNo.java
SwapnaWJun20: Above mentioned changes have been incorporated in the code.

src/swapnaWJun20/PalindromeString.java
SwapnaWJun20: Yes Anup. I got the point here. rectified the same

src/swapnaWJun20/PalindromeString.java
SwapnaWJun20: Done

src/swapnaWJun20/PalindromeNo.java
AnupSJun20: Now it looks good. For the math operation, it would be good to separate these by brackets. 
temp = (temp * 10 )+ (number1 % 10);

AnupSJun20 (approved):
code looks good now.

src/swapnaWJun20/PalindromeNo.java
```diff
@@ -0,0 +1,23 @@
...
+public class PalindromeNo {
+
+	void findPalindromeNumber(int number) {
+		int number1 = number;
+		int temp = 0;
+		while (number1 > 0) {
+			temp = temp * 10 + number1 % 10;
+			number1 = number1 / 10;
+		}
+		if (temp == number)
```
KrishnaKTechnocredits, line 12: Instead of temp you should use variable name as revNumber.

PR #666, opened 2020-07-01
Anup_Assignment_22_done_15min

AnupSJun20:
Anup_Assignment_22_done_15min. Please review.
Program1
Program2

### DIFF
--- a/src/anupSJun20/PalindromeNumber.java
+++ b/src/anupSJun20/PalindromeNumber.java
@@ -12,7 +12,7 @@ public class PalindromeNumber {
 		int initialNum = input;
 		int revNum = 0;
 		while (input > 0) {
-			revNum = revNum * 10 + input % 10;
+			revNum = (revNum * 10) + (input % 10);
 			input = input / 10;
 		}
 		if (initialNum == revNum)
--- a/src/anupSJun20/PalindromeNumber.java
+++ b/src/anupSJun20/PalindromeNumber.java
@@ -1,0 +1,28 @@
+package anupSJun20;
+/*Program 1: 
+Verify given number is palindrome or not. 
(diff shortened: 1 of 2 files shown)


src/anupSJun20/PalindromeNumber.java
```diff
@@ -0,0 +1,28 @@
...
+
+Input: number = 56789
+Output : Given number 56789 is not palindrome.*/
+
+public class PalindromeNumber {
+	static void findPalindromeNumber(int input) {
+		int initialNum = input;
+		int revNum = 0;
+		while (input > 0) {
+			revNum = revNum * 10 + input % 10;
```
harshvegada: Here put bracket for r(evNum * 10) + (input %10) so it will be more clear understand
Logic is perfectly placed

src/anupSJun20/PalindromeString.java
```diff
@@ -0,0 +1,26 @@
...
+Input: radar
+Output : Given string radar is palindrome.
+
+Input: technocredits
+Output: Given string technocredits is not palindrome.*/
+
+public class PalindromeString {
+	static void findPalindromeString(String str) {
+		String revstr = "";
+		for (int index = str.length() - 1; index >= 0; index--) {
```
harshvegada, line 13: here we are + operation so String + char will be also fine
Logically this is also good option

harshvegada (approved):
Everything looks good 
No comments on logic part 
Keep it up @AnupSJun20

src/anupSJun20/PalindromeNumber.java
AnupSJun20: Added bracket and pushed again.

PR #667, opened 2020-07-02
ADT_Assignment22_Added optimized solution for palindrome

AditiGJUN20:
@KrishnaKTechnocredits 
I've just added optimized solution for palindrome which you'd suggested in program for my future reference.

### DIFF
--- a/src/aditiGJUN20/PalindromeString.java
+++ b/src/aditiGJUN20/PalindromeString.java
@@ -9,20 +9,34 @@ package aditiGJUN20;
 
 public class PalindromeString {
 	boolean isPalindromString(String input) {
-		System.out.println("Input String: "+input);
-		String rev="";
-		for ( int index = input.length() - 1; index >= 0; index-- )
-        rev = rev + input.charAt(index); 
-		
-		if(input.equalsIgnoreCase(rev))
+		System.out.println("Input String: " + input);
+		String rev = "";
+		for (int index = input.length() - 1; index >= 0; index--)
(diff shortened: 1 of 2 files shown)


src/aditiGJUN20/PalindromeString.java
```diff
@@ -9,22 +9,36 @@
...
+//			}
+//		}
+//		if(isPalindrome)
+//			System.out.println(name+ " given string is palindrome.");
+//	}
+
 	public static void main(String[] args) {
 		PalindromeString palindromeString = new PalindromeString();
-		if(palindromeString.isPalindromString("Radar")==true)
+		if (palindromeString.isPalindromString("Radar") == true)
```
KrishnaKTechnocredits, line 39: == true comparision is not needed when you called method already return boolean value.

src/aditiGJUN20/PalindromeString.java
```diff
@@ -9,22 +9,36 @@
...
 public class PalindromeString {
 	boolean isPalindromString(String input) {
-		System.out.println("Input String: "+input);
-		String rev="";
-		for ( int index = input.length() - 1; index >= 0; index-- )
-        rev = rev + input.charAt(index); 
-		
-		if(input.equalsIgnoreCase(rev))
+		System.out.println("Input String: " + input);
+		String rev = "";
```
KrishnaKTechnocredits, line 13: rev not meaningful name, use revString.

PR #669, opened 2020-07-02
Aavruti_Assignment_23_5hrs

AavrutiDJun20:
Union and Intersection of two Array

UnionArray

IntersectionOfTwoArray

### DIFF
--- a/src/aavrutiDJun20/array/IntersectionOfTwoArray.java
+++ b/src/aavrutiDJun20/array/IntersectionOfTwoArray.java
@@ -1,0 +1,56 @@
+/*Find intersection(common elements) of two arrays.
+Input: arr1 -> {10.45, 14.0, 18.35, 88.88, 54.10, 18.35}
+          arr2 -> {17.20, 13.30, 10.45, 18.35, 84.33, 13.30}
+
+Output: 10.45, 18.35*/
+
+package aavrutiDJun20.array;
+
+public class IntersectionOfTwoArray {
+	int numberCnt = 0;
+	
+	double[] findIntersectionOfArray(double[] array1,double[] array2) {
+		double[] tempIntersectionArray = new double[array1.length+array2.length];
+		String[] str = new String[array1.length+array2.length];
+		str[0] = "";
+		for(int outerIndex=0;outerIndex<array1.length;outerIndex++) {
+			if(str[outerIndex]!= "*" ) {
+				for(int innerIndex=0;innerIndex<array2.length;innerIndex++) {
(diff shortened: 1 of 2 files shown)


src/aavrutiDJun20/array/IntersectionOfTwoArray.java
```diff
@@ -0,0 +1,74 @@
...
+		int numberCnt = 0;
+		int arrayLength = 0;
+		
+		if(array1.length>=array2.length)
+			arrayLength = array1.length;
+		else
+			arrayLength = array2.length;
+		
+		double[] tempIntersectionArray = new double[arrayLength];
+		String[] str = new String[arrayLength];
```
KrishnaKTechnocredits: @AavrutiDJun20 , String array is not required here and you haven't used it here.
at line number 31 you have if condition, you are checking * but you didnt write a logic to place a * at any index. isn't it ? According to me line number 17,18,31 is extra.

KrishnaKTechnocredits: I got it , you have used * logic but i think String array is an overhead and difficult to convince interviewer why we need String array for numbers

src/aavrutiDJun20/array/IntersectionOfTwoArray.java
```diff
@@ -0,0 +1,74 @@
...
+		for(int outerIndex=0;outerIndex<array1.length;outerIndex++) {			
+				for(int innerIndex=0;innerIndex<array2.length-1;innerIndex++) {
+					if(array1[outerIndex] == array2[innerIndex]) {
+						tempIntersectionArray[numberCnt] = array2[innerIndex];
+						numberCnt++;
+					}
+				}
+		}
+		
+		for(int outerIndex=0;outerIndex<numberCnt;outerIndex++) {
```
KrishnaKTechnocredits, line 28: Replacement of line number 29 to 44. I removed String array and use double array. 

double[] str = new double[arrayLength];
		for (int outerIndex = 0; outerIndex < numberCnt; outerIndex++) {
			boolean isNumberDuplicate = true;
			for (int innerIndex = 0; innerIndex < duplicateNumberCnt; innerIndex++) {
				if (tempIntersectionArray[outerIndex] == str[innerIndex]) {
					isNumberDuplicate = false;
					break;
				}
			}
			if (isNumberDuplicate == true) {
				str[duplicateNumberCnt] = tempIntersectionArray[outerIndex];
				duplicateNumberCnt++;
			}

		}
		return str;

KrishnaKTechnocredits (approved):
Code looks good to me. Keep it up.

PR #670, opened 2020-07-02
Aastha_Assignment_23_Done_Approx4hrs

AasthaJJun20:
->I have used a little different approach then suggested by Krishna (Though I will try to implement that as well)
->I have created an extra method which checks if the element is already present in given array
                  boolean searchInArray(double[] input, double element)
->This method is used by both programs(Union and intersection)
->PFB outputs:

1) Output when both arrays are of same length
ArraysUnionIntersection-bothEqualLength-output

2) Output  when arr1.length > arr2.length
ArraysUnionIntersection-arr1LengthLarge-output

3) Output  when arr1.length < arr2.length
ArraysUnionIntersection-arr2LengthLarge-output

### DIFF
--- a/src/aasthaJJun20/ArraysUnionIntersection.java
+++ b/src/aasthaJJun20/ArraysUnionIntersection.java
@@ -1,0 +1,80 @@
+package aasthaJJun20;
+
+public class ArraysUnionIntersection {
+	// This method is written just to display the array elements on console
+	private void displayArray(double[] input) {
+		System.out.print("{ ");
+		for (int index = 0; index < input.length; index++) {
+			System.out.print(input[index] + "   ");
+		}
+		System.out.println("}");
+	}
+
+	// Below method is written to search a particular element in the provided(input)
+	// array
+	private boolean searchInArray(double[] input, double element) {
+		for (int index = 0; index < input.length; index++) {
+			if (input[index] == element)
+				return true;
+		}
+		return false;
+	}
+
+	// Program 1: Find union of two arrays.
+	public double[] getUnionOfArray(double[] arr1, double[] arr2) {
+		double[] unionArray = new double[arr1.length + arr2.length];
+		int size = 0;
+		for (int index = 0; index < arr1.length; index++) {
+			if (searchInArray(unionArray, arr1[index]) == false) {
+				unionArray[size] = arr1[index];
+				size++;
+			}
+		}
+		for (int index = 0; index < arr2.length; index++) {
+			if (searchInArray(unionArray, arr2[index]) == false) {
+				unionArray[size] = arr2[index];
+				size++;
+			}
+		}
+		double[] unionArrayResult = new double[size];
+		for (int index = 0; index < size; index++) {
+			unionArrayResult[index] = unionArray[index];
+		}
+		return unionArrayResult;
+	}
+
+	// Program 2: Find intersection(common elements) of two arrays.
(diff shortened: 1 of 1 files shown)


src/aasthaJJun20/ArraysUnionIntersection.java
```diff
@@ -0,0 +1,80 @@
...
+		}
+		return false;
+	}
+
+	// Program 1: Find union of two arrays.
+	public double[] getUnionOfArray(double[] arr1, double[] arr2) {
+		double[] unionArray = new double[arr1.length + arr2.length];
+		int size = 0;
+		for (int index = 0; index < arr1.length; index++) {
+			if (searchInArray(unionArray, arr1[index]) == false) {
```
KrishnaKTechnocredits, line 28: searchArray length is sum of two array length, hence for loop should be iterated over searchArray length time to search every element. we can use little trick here to reduce number of comparison.

// Below method is written to search a particular element in the provided(input)
	// array
	private boolean searchInArray(double[] input, double element, int size) {
		for (int index = 0; index < size; index++) {
			if (input[index] == element)
				return true;
		}
		return false;
	}

src/aasthaJJun20/ArraysUnionIntersection.java
```diff
@@ -0,0 +1,80 @@
...
+				unionArray[size] = arr1[index];
+				size++;
+			}
+		}
+		for (int index = 0; index < arr2.length; index++) {
+			if (searchInArray(unionArray, arr2[index]) == false) {
```
KrishnaKTechnocredits, line 34: if (searchInArray(unionArray, arr1[index],size) == false) {

pass size variable here, why do we keep compare all values.

src/aasthaJJun20/ArraysUnionIntersection.java
```diff
@@ -0,0 +1,80 @@
...
+	public double[] getIntersectionOfArray(double[] arr1, double[] arr2) {
+		int size = 0;
+		if (arr1.length > arr2.length)
+			size = arr1.length;
+		else
+			size = arr2.length;
+		double[] intersectionArray = new double[size];
+		size = 0;
+		for (int index = 0; index < arr2.length; index++) {
+			if (searchInArray(arr1, arr2[index])) {
```
KrishnaKTechnocredits, line 56: if (searchInArray(unionArray, arr1[index],arr1.length-index) == false) {

think about it :)

src/aasthaJJun20/ArraysUnionIntersection.java
AasthaJJun20: Yes, Krishna I was thinking of it, that how can I reduce the number of iterations, this is the perfect solution to it.
Thanks a lot. Implementing it now :)

src/aasthaJJun20/ArraysUnionIntersection.java
AasthaJJun20: Here, When I'm passing **arr1.length-index** parameter, it is leaving the last element of arr2(which is common in both arrays).
when tried with just **arr1.length**, the output looks fine.

KrishnaKTechnocredits: Thats correct, you got it.
We can think of minimizing iteration in case of intersection. Here you used same method in both cases (thats actually nice) so have to think generalized solution.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #672, opened 2020-07-02
Mahesh_Assignment_23_(60+30)mins_Done

MaheshKJun20:
ArrayUnion
ArrayUnionUsingString
ArrayIntersection
Mahesh_Assignment_23_(60+30)mins_Done


src/maheshKJun20/ArrayIntersection.java
```diff
@@ -0,0 +1,68 @@
...
+	int count=0;
+	void findArrayIntersection(double[] arr1, double[] arr2) {
+		double[] intersectionOfArray = new double[arr1.length];
+		int counter =0;
+		for(int index=0;index<arr1.length;index++) {
+			for(int innerIndex=0;innerIndex<arr2.length;innerIndex++) {
+				if(arr2[innerIndex] == arr1[index]) {
+					intersectionOfArray[counter] = arr1[index];
+					counter++;
+				}
```
harshvegada: Here after counter++, you can put break as you already get the index value. why go and check further elements and update second array value to any other value so if same element from array again come and comparing it will not comes into if condition as we already process same element.

image

harshvegada: this change save your findUniqueValueArray() method calling n comparing part.

src/maheshKJun20/ArrayExample/ArrayUnionUsingString.java
```diff
@@ -0,0 +1,63 @@
+package maheshKJun20.ArrayExample;
+
+public class ArrayUnionUsingString {
```
harshvegada, line 3: Not matching the expectation.

harshvegada (commented):
Please work the comments

In this ArrayIntersection class getting expected output add comments for improving code,
and remove the not code which not match the expectation. 

Please remove the ArrayUnionUsingString.java as its duplicate file of ArrayUnion.java

src/maheshKJun20/ArrayIntersection.java
MaheshKJun20: Thanks Harsh for the changes Suggested. I have updated and commited the code as oer above suggested changes.

KrishnaKTechnocredits (approved):
I am okay with Logic but I feel, it should be without String array. Please take a look at Aastha's code. She managed it well.

PR #675, opened 2020-07-03
Anshu-Assignment-23-150min+45min-Done

AnshuKJun20:
image
image

### DIFF
--- a/src/anshuKJun20/IntersectionOfArray.java
+++ b/src/anshuKJun20/IntersectionOfArray.java
@@ -1,0 +1,44 @@
+package anshuKJun20;
+
+import java.util.Arrays;
+
+public class IntersectionOfArray {
+	
+	double[] getIntersectionArray(double[] arr1, double[] arr2) {
+		double[] tempArr = new double[arr2.length];
+		int count = 0;
+		for (int outerindex = 0; outerindex < arr2.length; outerindex++) {
+			boolean flag = false;
+			for (int innerindex = 0; innerindex < arr1.length; innerindex++) {
+				if (arr2[outerindex] == arr1[innerindex]) {
+					flag = true;
+					break;
+				}
+			}
+			if (flag == true) {
+				tempArr[count] = arr2[outerindex];
+				count++;
+			}
+		}
+		double[] intersectionArray = new double[count];
+		for (int index = 0; index < count; index++) {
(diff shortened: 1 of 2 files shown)


src/anshuKJun20/IntersectionOfArray.java
```diff
@@ -0,0 +1,44 @@
...
+		int count = 0;
+		for (int outerindex = 0; outerindex < arr2.length; outerindex++) {
+			boolean flag = false;
+			for (int innerindex = 0; innerindex < arr1.length; innerindex++) {
+				if (arr2[outerindex] == arr1[innerindex]) {
+					flag = true;
+					break;
+				}
+			}
+			if (flag == true) {
```
PranitaPJun20: 1.Initially you set flag =false.In first if statement you are changing flag value to 'true' and again you are checking for 
if (flag == true) condition.This not correct.
2.Actually in this code you dont need to set flag value.Without setting also you can get desired output.
3.Line number 19 and 20 merge in first If statement and delete second if statement.

PranitaPJun20 (commented):
Please work on given comments.

PranitaPJun20 (approved):
Corrected as per comments.

src/anshuKJun20/IntersectionOfArray.java
AnshuKJun20: You are right, No need of flag here. Changes Done.

src/anshuKJun20/IntersectionOfArray.java
```diff
@@ -0,0 +1,45 @@
...
+			double[] tempArr = new double[arr1.length];
+		}else {
+			double[] tempArr = new double[arr2.length];
+		}
+		int count = 0;
+		for (int outerindex = 0; outerindex < arr2.length; outerindex++) {
+			
+			for (int innerindex = 0; innerindex < arr1.length; innerindex++) {
+				if (arr2[outerindex] == arr1[innerindex]) {
+					tempArr[count] = arr2[outerindex];
```
KrishnaKTechnocredits, line 18: I think here it should be a compilation error. Please verify.

AnshuKJun20: image

I checked it and also debug it. There is no any compilation error.

AnshuKJun20: image

PR #676, opened 2020-07-03
Viresh_Assignment_23_Done_160_Mins

VireshJJun20:
@Krishna, please review and comment.
image

image

### DIFF
--- a/src/vireshJJun20/Assignment23/ArrayIntersection.java
+++ b/src/vireshJJun20/Assignment23/ArrayIntersection.java
@@ -1,0 +1,86 @@
+/*
+Program 2: Find intersection(common elements) of two arrays.
+Input:    arr1 -> {10.45, 14.0, 18.35, 88.88, 54.10, 18.35}
+          arr2 -> {17.20, 13.30, 10.45, 18.35, 84.33, 13.30}
+
+Output: 10.45, 18.35
+Method Signature : double[] getIntersectionOfArray(double[] arr1, double[] arr2)
+*/
+
+
+package vireshJJun20.Assignment23;
+
+public class ArrayIntersection {
+
+	static int count;
+	boolean flag = false;
+
+	double[] editArray(double[] ar) { // takes input array and removes duplicates and then returns.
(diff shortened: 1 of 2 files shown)


src/vireshJJun20/Assignment23/ArrayIntersection.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class ArrayIntersection {
+
+	static int count;
+	boolean flag = false;
+
+	double[] editArray(double[] ar) { // takes input array and removes duplicates and then returns.
+		double[] temp = new double[ar.length];
+		for (int outerIndex = 0; outerIndex < ar.length; outerIndex++) {
+			for (int innerIndex = 0; innerIndex < ar.length; innerIndex++) {
```
KrishnaKTechnocredits, line 21: Loop is iterating for more number of time and line number 22 to 27 can be written in better way.

for (int outerIndex = 0; outerIndex < arr1.length; outerIndex++) {
			boolean flag = true;
			for (int innerIndex = 0; innerIndex < count; innerIndex++) {
				if (arr1[outerIndex] == tempAr[innerIndex]) {
					flag = false;
					break;
				}
			}
			if (flag == true) {
				tempAr[count] = arr1[outerIndex];
				count++;
			}
		}

I wrote it on the fly to give you an idea. please verify and revert back if you agree.

VireshJJun20: Hi Krishna, I have observed your code snippet, and made changes in innerloop (innerIndex < count). if we skip the else condition (else {
					flag = true;
					break;
				}), 
then the Arr1 duplicate value remains as it is in a resultant array.  and the output reflects as below.
image

In case if the else condition is present, then output reflects correctly. As below:

image

VireshJJun20: And lastly if I change the if ()condition, then the output is blank. as highlighted below:

image

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #678, opened 2020-07-03
Raj_Assignment_23_Done

RajAJun20:
Assignment_23_P1
Assignment_23_P2

### DIFF
--- a/src/rajAJun20/IntersectionOfArrays.java
+++ b/src/rajAJun20/IntersectionOfArrays.java
@@ -1,0 +1,51 @@
+package rajAJun20;
+
+public class IntersectionOfArrays {
+	int countTemp = 0;
+	
+	// Common numbers
+	void getIntersectionOfArray(double[] array1, double[] array2) {
+		double[] tempIntersectionArray = new double[(array1.length)];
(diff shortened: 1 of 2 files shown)


src/rajAJun20/IntersectionOfArrays.java
```diff
@@ -0,0 +1,51 @@
+package rajAJun20;
+
+public class IntersectionOfArrays {
+	int countTemp = 0;
+	
+	// Common numbers
+	void getIntersectionOfArray(double[] array1, double[] array2) {
+		double[] tempIntersectionArray = new double[(array1.length)];
```
AnshuKJun20, line 8: Here you have to first check which array has minimum length, then take  minimum length is length of your tempIntersectionArray[].

AnshuKJun20 (commented):
Please work on comments.

src/rajAJun20/IntersectionOfArrays.java
KrishnaKTechnocredits: @RajAJun20 I am agree with Anshu's comment. Good job @AnshuKJun20

KrishnaKTechnocredits (approved):
Apart from Array length, Code looks good to me.

PR #682, opened 2020-07-03
Archana_Assignment_23_Done_2Hrs

ArchanaNJun20:
ArrayUnion:
image
ArrayIntersection:
image

### DIFF
--- a/src/archanaNJun20/ArrayIntersection.java
+++ b/src/archanaNJun20/ArrayIntersection.java
@@ -4,7 +4,7 @@ public class ArrayIntersection {
 	double[] getIntersectionOfArray(double[] array1, double[] array2){		
 		int  arraycount=0;
 		int length=0;
-		if(array1.length>array2.length)
+		if(array1.length<array2.length)
 			length= array1.length;
 		else
 			length=array2.length;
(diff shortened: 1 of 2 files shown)


src/archanaNJun20/ArrayIntersection.java
```diff
@@ -0,0 +1,43 @@
+package archanaNJun20;
+
+public class ArrayIntersection {
+	double[] getIntersectionOfArray(double[] array1, double[] array2){		
+		int  arraycount=0;
+		int length=0;
+		if(array1.length>array2.length)
+			length= array1.length;
```
KrishnaKTechnocredits, line 8: For intersection of an array, you should create array length of min array. You have written logic for max array. Change it.

src/archanaNJun20/ArrayUnion.java
```diff
@@ -0,0 +1,59 @@
+package archanaNJun20;
+
+public class ArrayUnion {
+	int arraycount=0;
+	double[] getUnionOfArray(double[] array1,double[] array2){
+		int arraycount=0;
```
KrishnaKTechnocredits, line 5: Why you have created two variable with same name? Any specific reason?

ArchanaNJun20: sorry, my bad.by mistake I declare it twice.

src/archanaNJun20/ArrayIntersection.java
ArchanaNJun20: The code is updated.
Thank you :)

PR #688, opened 2020-07-04
Kartikey_Assignment_23_2+3hrs

KartikeyDJune20:
image
image

### DIFF
--- a/src/kartikeyDJune20/InetrsectionOfArrays.java
+++ b/src/kartikeyDJune20/InetrsectionOfArrays.java
@@ -1,0 +1,59 @@
+package kartikeyDJune20;
+
+import java.util.Arrays;
+
+public class InetrsectionOfArrays {
+	static void intersectionArray(double[] arrInput1, double[] arrInput2) {
+		int lenTempArray = 0;
+		if (arrInput1.length > arrInput2.length)
+			lenTempArray = arrInput2.length;
+		else
+			lenTempArray = arrInput1.length;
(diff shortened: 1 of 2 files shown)


src/kartikeyDJune20/InetrsectionOfArrays.java
```diff
@@ -0,0 +1,59 @@
...
+		int lenTempArray = 0;
+		if (arrInput1.length > arrInput2.length)
+			lenTempArray = arrInput2.length;
+		else
+			lenTempArray = arrInput1.length;
+
+		double[] tempIntersectionArry = new double[lenTempArray];
+		int cnt = 0;
+		Arrays.sort(arrInput1);
+		Arrays.sort(arrInput2);
```
KrishnaKTechnocredits: Why we are using this method?  Remove it and implement it without it.

src/kartikeyDJune20/UnionOfArrays.java
```diff
@@ -0,0 +1,63 @@
+package kartikeyDJune20;
+
+import java.util.Arrays;
+
+public class UnionOfArrays {
+
+	static void unionOfArray(double[] arrInput1, double[] arrInput2) {
+		double[] tempUnionArray = new double[arrInput1.length + arrInput2.length];
+		Arrays.sort(arrInput1);
```
KrishnaKTechnocredits: Don't use it.

PR #691, opened 2020-07-05
Mahesh_Coding_Exam_4_5Mins

MaheshKJun20:
image


harshvegada (approved):
Now, working as expected @MaheshKJun20 
Do not hurry! check for other inputs

PR #692, opened 2020-07-05
Karan_Coding_Exam4_20mins

KARAN1595:
completed


harshvegada (approved):
Good job @KARAN1595 
No Comments

two suggestion on code:

1. One   correction if you already initialised largest as a[0] then why need to start   for loop from 0 index, start from 1     
2. In else if condition you should add one more condition as currectNumber   !=maxNumber to prevent risk that max and sMax are same

PR #693, opened 2020-07-05
Archana_coding_Exam_4_10mins

ArchanaNJun20:
output:
image


harshvegada (approved):
Do not take firstmax n second max as 0
this works for positive array but for negative it will fail.

PR #694, opened 2020-07-05
Ajit_CodingExam_4_Completed_25mins

AjitSJune20:
image

We have two options to initialize  variable firsthighest, secondhighest
firsthighest =array[0] ,  secondhighest = array[0] , then Start loop from Index 1

but I tried to implement other concept  firsthighest =Integer.MIN_VALUE ,  secondhighest =Integer.MIN_VALUE


harshvegada (approved):
Excellent work @AjitSJune20 
No Comments everything as expected

PR #695, opened 2020-07-05
Piyush_Coding_Exam_4_30Min

PiyushPJun20:
image

### DIFF
--- a/src/SecondMax.java
+++ b/src/SecondMax.java
@@ -1,0 +1,28 @@
+public class SecondMax {
+
+	void displaySecondMax(int[] number) {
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
No Comments on logic part but two suggestion on code part

1. One correction if you already initialised largest as a[0] then why need to start for loop from 0 index, start from 1
2. In else if condition you should add one more condition as currectNumber !=maxNumber to prevent risk that max and sMax are same

harshvegada (approved):
Logically correct output but, One correction if  you already initialised largest as a[0] then why need to start for loop from  0 index

PR #696, opened 2020-07-05
Pranita_Coding_Exam_4_done_20mins

PranitaPJun20:
image


harshvegada (approved):
@PranitaPJun20 Perfectly placed code,
No Comments
Keep it up!
One Suggestion:

One correction if you already initialised largest as a[0] then why need to start for loop from 0 index, start from 1

PR #697, opened 2020-07-05
AMITA_Coding_Exam_4_Done_25mins

AmitaRJun20:
AMITA_Coding_Exam_4_Done...please review
image


harshvegada (approved):
correctly place logic, excellent work @AmitaRJun20 
Always different than others

One suggestion:
on second If condition you should add one more condition as currectNumber !=maxNumber to prevent risk that max and sMax are same

PR #698, opened 2020-07-05
Kartikey_Coding_Exam_4_10-15mins

KartikeyDJune20:
image


harshvegada (approved):
Excellent work @KartikeyDJune20 
Keep it up
One correction in printing, you need to print second highes, but logic is correct for second large number

PR #699, opened 2020-07-05
Deavina_Coding_Exam_4_Done_20mins

DeavinaSJun20:
image


harshvegada (approved):
No Comments on logic part
One correction if   you already initialised largest as a[0] then why need to start for loop from   0 index

PR #700, opened 2020-07-05
ADT_Exam_4_17Mins

AditiGJUN20:
Exam 4:
Time: 17 Mins
Screen Shot 2020-07-04 at 11 06 42 PM

Extra commit - I've moved condition to check invalid input from main to method logic.


harshvegada (approved):
Good attempt @AditiGJUN20 
Different than others logic, two corrections

1. on iteration, you define firstMax as a[0] and secondMax as a[0] and you start loop from 0, that not good poing, no logical impact just one iteration extra.
2. In else if condition you should add one more condition as currectNumber !=maxNumber to prevent risk that max and sMax are same

PR #701, opened 2020-07-05
Jagadeesh_Coding_Exam_4_20min

JagadeeshMJun20:
Please review and comment. 
Output:
image


harshvegada (approved):
Logically Correct but never take highest n second Highest number 0 take always from array.

PR #702, opened 2020-07-05
Aastha_Coding_exam_20mins

AasthaJJun20:
PFB screenshot of output:
image


harshvegada (approved):
Perfectly Placed logic @AasthaJJun20 
Keep it up!
M expecting something like this
One correction in code
1. In else if condition you should add one more condition as currectNumber !=maxNumber to prevent risk that max and sMax are same

PR #703, opened 2020-07-05
Rachana_Coading_Exam_4_20 min

RachanaGJun20:
image


harshvegada (approved):
One   correction if you already initialised largest as a[0] then why need to start   for loop from 0 index

RachanaGJun20:
> One correction if you already initialised largest as a[0] then why need to start for loop from 0 index

oh Yes got it Harsh... Thank you . will do change it.

PR #704, opened 2020-07-05
Suparna_Coding_exam04_15Min

SuparnaNikam:
SecondMazNo

### DIFF
--- a/src/suparnaNJune20/CodingExam/SecondMaxNo.java
+++ b/src/suparnaNJune20/CodingExam/SecondMaxNo.java
@@ -1,4 +1,4 @@
-
+package suparnaNJune20.CodingExam;
 public class SecondMaxNo {
 
 	void fndSecondMaxNo(int number[]) {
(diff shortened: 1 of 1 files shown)


src/suparnaNJune20/CodingExam/SecondMaxNo.java
```diff
@@ -0,0 +1,21 @@
+package suparnaNJune20.CodingExam;
+public class SecondMaxNo {
+
+	void fndSecondMaxNo(int number[]) {
+		int max1No = number[0], max2No = number[0];
+		for (int outerIndex = 0; outerIndex < number.length; outerIndex++) {
+			if (number[outerIndex] > max1No && number[outerIndex] > max2No) {
```
harshvegada, line 7: here don;t need to add condition for max2No, because anyways if current number is greater than max number then it is always greater than secondmax number

harshvegada (approved):
logically correct, work on comment next time

If current number   greater than max number than obliviously current number always greater than   secondMax
--

PR #705, opened 2020-07-05
Madhura_Coding_Exam_4_30mins

MadhuraMJun20:
image


harshvegada (approved):
Very well @MadhuraMJun20 
I guess this is bubble sort algorithm, as we talk in class this take this will take double Number of complexity 
our work will work will done in single for loop as well. 
Improve code next time, don't go for algorithms 


hint:
if(currentNumber>max){
   //swap sMax and max logic
}else if(currentNumber!=max && currentNum>sMax){
   //change sMax value
}

try this code in one loop

PR #706, opened 2020-07-05
Swapna_Coding_Exam_4_Done_30_min

SwapnaWJun20:
image


harshvegada (approved):
logically correct but 
One correction if   you already initialised largest as a[0] then why need to start for loop from   0 index

PR #707, opened 2020-07-05
Abhishek_codingExam_4_35min

AbhishekSJune2020:
Second Maximum number is: 56


harshvegada (approved):
NO logically change but One correction if   you already initialised largest as a[0] then why need to start for loop from   0 index

PR #708, opened 2020-07-05
Apurva_Test4_Done_30mins

ApurvaJun20:
image


harshvegada (commented):
Good job @ApurvaJun20 
Can you explain why we are passing array length from main method, that is something you can handle in you service method right?
I see very same code on google
https://www.geeksforgeeks.org/find-second-largest-element-array/

PR #709, opened 2020-07-05
Raj_Coding_Exam_4_done_28mins

RajAJun20:
Test_4


harshvegada (approved):
Good job @RajAJun20 
Logically correct!
But, One correction if   you already initialised largest as a[0] then why need to start for loop from   0 index

PR #710, opened 2020-07-05
Vaishnavi_Coding_Exam_Done_50Min_Done

VaishnaviVJun20:
image


harshvegada (approved):
Good Job @VaishnaviVJun20 Logically correct one suggestion in code is Never take input as   0 always take input from array

PR #711, opened 2020-07-05
Nikhil_Coding_Exam_4_30Min

NikhilMJun20:
->Coding Exam 4 completed and pushed.
->Completed in 5 min only but, using 2 for loops.
->Was trying to find with one for loop but, could not do it.
->Please review and comment.

->PFB screenshot of the output:-
Annotation 2020-07-05 120427


harshvegada (approved):
Good job @NikhilMJun20 
Looks like this bubble sort algorithm, as we talk this will take two for loop to get answer, complexity is double,
Think on one for loop, can we manage this in one loop

hint:
if(currentNumber>max){
   //swap sMax and max logic
}else if(currentNumber!=max && currentNum>sMax){
   //change sMax value
}

PR #714, opened 2020-07-05
Aavruti_Coding_Exam_4_50mins

AavrutiDJun20:
Exam_4


harshvegada (approved):
Good job @AavrutiDJun20 
Keep it up!

One correction if   you already initialised largest as a[0] then why need to start for loop from   0 index, start from 1

PR #715, opened 2020-07-05
Barkha_CodingTest4_45mins

BarkhaPJun20:
Find Second Highest Number in Array:


harshvegada (approved):
Code is correct but  two improvements

1. in else if condition you should add one more condition as currectNumber !=maxNumber 
2. Never take input as   0 always take input from array

PR #716, opened 2020-07-05
Viresh_Coding_Exam_4_Done_45Mins

VireshJJun20:
Please review and comment.

image


harshvegada (approved):
Code is correct here the two suggestions


1. in   else if condition you should add one more condition as currectNumber   !=maxNumber     
2. Never take input as 0 always take input from array

PR #717, opened 2020-07-05
Shruti Coding Exam 45 mints

ShrutiDJun20:


harshvegada (approved):
Code is correct @ShrutiDJun20 but two suggestions

1. in else if condition you should add one more condition as currectNumber !=maxNumber to prevent risk that max and sMax are same
2. Never take input as 0 always take input from array

PR #718, opened 2020-07-05
Anup_Coding_Exam_4_Done_40mins

AnupSJun20:
Please review.
Program1


harshvegada (approved):
Code looks correct @AnupSJun20 
One suggestion

1. in else if condition you should add one more condition as currectNumber !=maxNumber to prevent risk that max and sMax are same
2. One correction if you already initialised largest as a[0] then why need to start for loop from 0 index, start from 1

PR #719, opened 2020-07-05
Palak_Coding_Exam4_45min

PalakSJun20:
Firstly I was working with two loops and  was unable to get results 
image


harshvegada (approved):
Good job @PalakSJun20 
two thing i want to add in code part

1. Never take input as 0 always take input from array
2. in else if condition you should add one more condition as currectNumber !=maxNumber to prevent risk that max and sMax are same

rest logic correct

PR #725, opened 2020-07-06
Anshu-Assignment-24-1hr-Done

AnshuKJun20:
image
image


PranitaPJun20 (approved):
The output is exactly as expected.

harshvegada (approved):
Code looks good @AnshuKJun20 
you don't need to mention SOP("enter number like ******* format just SOP("INVALID")")

PR #726, opened 2020-07-06
Barkha_Assignment_24_45mins

BarkhaPJun20:
Validate given phone number.
+91 9765463742 --> valid
09765463742 --> valid
976 546 3742 --> valid
9765463742 --> valid
9765463A42 --> Invalid
976546342 --> Invalid
6976546342 --> Invalid
image

### DIFF
--- a/src/barkhaPJun20/VerifyMobileNumber.java
+++ b/src/barkhaPJun20/VerifyMobileNumber.java
@@ -1,77 +1,54 @@
 package barkhaPJun20;
 
 public class VerifyMobileNumber {
-	
+	private boolean isNumber(String num){
+		for (int index = 0; index < num.length(); index++) {
+			if (Character.isAlphabetic(num.charAt(index)))
+				return false;                                 // if return false then not a number.(include alphabet)
+		}
+		return true;
+	}
 	void validMobileNumber(String num) {
+		System.out.println("Mobile Number provided to validate is: "+num);
 		//Exa.+91 9765463742
 		if (num.length()==14) {
-			if (num.startsWith("+91")&& num.charAt(4)=='9'){
-				boolean flag=true;
-				for (int i=4;i<num.length();i++){
-					if (Character.isAlphabetic(num.charAt(i))){
-						flag=false;
-						break;
-					}
-				}
-				if (flag)
+			if (num.startsWith("+91")&& num.charAt(4)=='9' && isNumber(num.substring(4)) && isNumber(num)) {
+				
 					System.out.println("Number is Valid");
-				else 
+			} else 
 					System.out.println("Number is Invalid");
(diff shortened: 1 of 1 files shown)


src/barkhaPJun20/VerifyMobileNumber.java
```diff
@@ -0,0 +1,77 @@
+package barkhaPJun20;
+
+public class VerifyMobileNumber {
+	
+	void validMobileNumber(String num) {
+		//Exa.+91 9765463742
+		if (num.length()==14) {
```
SwapnaWJun20, line 14: Here, you can write both if conditions together by using && operator.

src/barkhaPJun20/VerifyMobileNumber.java
```diff
@@ -0,0 +1,77 @@
...
+		}else 
+		System.out.println("Number is Invalid");
+	}	
+	//out of Desire Length
+	if (num.length()>14  || num.length()<10)
+		System.out.println("Number is Invalid");
+	}	
+	public static void main(String[] args) {
+		VerifyMobileNumber  mobileNumber=new VerifyMobileNumber ();
+		String mobileNum="9765463A42";
```
SwapnaWJun20: Mob no can be directly added while calling method.

src/barkhaPJun20/VerifyMobileNumber.java
BarkhaPJun20: Changes Done

src/barkhaPJun20/VerifyMobileNumber.java
BarkhaPJun20: Changes Done

src/barkhaPJun20/VerifyMobileNumber.java
```diff
@@ -0,0 +1,54 @@
...
+		//out of Desire Length
+		else 
+			System.out.println("Number is Invalid");
+		}	
+public static void main(String[] args) {
+		VerifyMobileNumber  mobileNumber=new VerifyMobileNumber ();
+		//String mobileNum="+91 9765463742";
+		//System.out.println("Mobile Number provided to validate is: "+mobileNum);
+		mobileNumber.validMobileNumber("+91 8565463742");
+	}
```
SwapnaWJun20, line 52: Please try to check 979 567894 this no as input, this is not giving proper output. As it has space in between and the length is 10.

BarkhaPJun20: Changes Done

SwapnaWJun20 (approved):

SwapnaWJun20:
Also for future reference, you can use single if loop with && and || operators.

PR #730, opened 2020-07-06
Swapna_Assignment_24_Suggested_Changes_Done_30_Min

SwapnaWJun20:
image

### DIFF
--- a/src/swapnaWJun20/MobileNoValidity.java
+++ b/src/swapnaWJun20/MobileNoValidity.java
@@ -1,0 +1,36 @@
+package swapnaWJun20;
+
+public class MobileNoValidity {
+
+	void checkMobileNo(String mobNo) {
+		if (mobNo.length() > 14 || mobNo.length() < 10) {
+			System.out.println("\n" + mobNo + " : Mobile no is invalid.");
+		} else {
+			if (mobNo.length() == 10 && mobNo.startsWith("9") || mobNo.startsWith("8") || mobNo.startsWith("7")) {
+				System.out.println("\n" + mobNo + " : Mobile no is valid.");
+			} else if (mobNo.length() == 11 && mobNo.startsWith("0")) {
+				System.out.println("\n" + mobNo + ": Mobile no is valid.");
+			} else if (mobNo.length() == 12 && mobNo.charAt(3) == ' ' && mobNo.charAt(7) == ' ') {
+				System.out.println("\n" + mobNo + ": Mobile no is valid.");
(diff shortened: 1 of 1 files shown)


src/swapnaWJun20/MobileNoValidity.java
```diff
@@ -0,0 +1,64 @@
...
+		}
+		return flag;
+	}
+
+	void checkMobileNo(String mobNo) {
+		if (mobNo.length() > 14 || mobNo.length() < 10) {
+			System.out.println("\n" + mobNo + " : Mobile no is invalid.");
+		} else {
+			if (mobNo.length() == 10 && mobNo.startsWith("9") || mobNo.startsWith("8") || mobNo.startsWith("7")) {
+				checkAlphabet(mobNo);
```
AnupSJun20: Here, in the If condition, we can use directly the method calling as you are returning the value and using that. One checkAlphabet(mobNo) calling and if(flag == true) not required together.

src/swapnaWJun20/MobileNoValidity.java
```diff
@@ -0,0 +1,64 @@
...
+		mobileNoValidity.checkMobileNo("09765463742");
+		mobileNoValidity.checkMobileNo("976 546 3742");
+		mobileNoValidity.checkMobileNo("9765463742");
+
+		mobileNoValidity.checkMobileNo("998979765463742"); // invalid
+		mobileNoValidity.checkMobileNo("0765463742");// invalid
+		mobileNoValidity.checkMobileNo("99897  83742"); // invalid
+		mobileNoValidity.checkMobileNo("998 97 83742");// invalid
+		mobileNoValidity.checkMobileNo("998 979 S7Y2");// invalid
+	}
```
AnupSJun20, line 63: Please enter a number with 979 567894, although it is 9 digits and with a space it becomes 10 as the length, so the code returns this as a valid number. Please add the space criteria.
Rest you have covered all the conditions well.

AnupSJun20 (changes requested):
Please incorporate the changes. Covered almost all scenarios.

src/swapnaWJun20/MobileNoValidity.java
SwapnaWJun20: I have added the check specified above to the code.

src/swapnaWJun20/MobileNoValidity.java
SwapnaWJun20: Updated the code as per the comments.

AnupSJun20 (approved):
Thanks for adding the validation. Code looks good. Approving.

PR #731, opened 2020-07-06
Aastha_Assignment_24_Done_30mins

AasthaJJun20:
->Created 2 utility methods : isNumber  and startOfNumber
->isNumber  checks if the given mobile contains any alphabet/special char
->startOfNumber checks if the given number starts with 9/8/7
PFB screenshot:
ValidateMobileNumber-output

### DIFF
--- a/src/aasthaJJun20/ValidateMobileNumber.java
+++ b/src/aasthaJJun20/ValidateMobileNumber.java
@@ -1,0 +1,67 @@
+package aasthaJJun20;
+
+//Assignment-24 - Validate given phone number 
+public class ValidateMobileNumber {
+	// below method checks if there is any alphabet or special char in given number
+	private boolean isNumber(String number) {
+		for (int index = 0; index < number.length(); index++) {
+			if (!Character.isDigit(number.charAt(index)))
+				return false;
+		}
+		return true;
+	}
+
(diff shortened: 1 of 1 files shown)


src/aasthaJJun20/ValidateMobileNumber.java
```diff
@@ -0,0 +1,67 @@
...
+			return true;
+		return false;
+	}
+
+	public void verifyMobileNumber(String number) {
+		switch (number.length()) {
+		// below case - is to validate numbers like : +91 9765463742
+		case 14:
+			if ((number.startsWith("+91 ")) && (checkFirstDigit(number.charAt(4))) && isNumber(number.substring(4)))
+				System.out.println(number + " --> Valid");
```
NikhilMJun20, line 26: One suggestion: By using substring, the input string is sanitized (i.e. getting rid of  + here) and then given to isNumber method. We need to give different substrings for different types of Ph No strings also, we need to remove spaces in another format. This is logically absolutely fine but, I would recommend we do this sanitization in the method and ask user to pass just the string he has...

NikhilMJun20 (approved):
Logic of the code looks good to me.
I am approving the PR.

KrishnaKTechnocredits (approved):
Good job Aastha.

PR #733, opened 2020-07-06
Vaishnavi_Assignment_24_30Min_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20_arrayExamples/ValidateGivenPhone.java
+++ b/src/vaishnaviVJun20_arrayExamples/ValidateGivenPhone.java
@@ -1,0 +1,46 @@
+package vaishnaviVJun20_arrayExamples;
+
+/*Validate given phone number.
+
++91 9765463742 --> valid
+09765463742 --> valid
+976 546 3742 --> valid
+9765463742 --> valid
+
+9765463A42 --> Invalid
+976546342 --> Invalid
+6976546342 --> Invalid
+
+Note : phone number is given in String format. */
+
+public class ValidateGivenPhone {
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Correct Understanding @VaishnaviVJun20 
Keep it up!
No Comments

One point i wanted to add is make it pieces so it will be easy to understand.

src/vaishnaviVJun20_arrayExamples/ValidateGivenPhone.java
```diff
@@ -0,0 +1,46 @@
...
+Note : phone number is given in String format. */
+
+public class ValidateGivenPhone {
+	int startsWith;
+	boolean flag = true;
+	
+	void validateGivenPhoneNumber(String[] array) {
+		for(int index=0;index<array.length;index++) {
+			if(array[index].length() == 14 || array[index].length() == 12 || array[index].length() == 11 || array[index].length() ==10) {
+				if(array[index].substring(0,4).equals("+91 ") || array[index].charAt(startsWith) == '0'   || array[index].charAt(startsWith) == '9' || array[index].charAt(startsWith) == '7' || array[index].charAt(startsWith) == '8') {
```
harshvegada: using this approach, number of lines will be less, readability will be less.
so make this statement in switch case (length) and then make method which check other verification point for mobile number. make as mush as modular code

VaishnaviVJun20: - Thank you for the suggestion.
- Done suggested changes.

PR #742, opened 2020-07-07
Archana_Assignment_24_Done_45min

ArchanaNJun20:
image

### DIFF
--- a/src/archanaNJun20/PhoneNumberValidation.java
+++ b/src/archanaNJun20/PhoneNumberValidation.java
@@ -69,27 +69,10 @@ public class PhoneNumberValidation {
 	}
 	public static void main(String[] args) {
 		PhoneNumberValidation phoneNumberValidation=new PhoneNumberValidation();
-		System.out.println("phone number of length :10");
-		phoneNumberValidation.checkPhoneNumberValidation("9028553461");
-		phoneNumberValidation.checkPhoneNumberValidation("90 8553461");
-		phoneNumberValidation.checkPhoneNumberValidation("90y8553461");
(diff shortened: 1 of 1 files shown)


src/archanaNJun20/PhoneNumberValidation.java
```diff
@@ -0,0 +1,95 @@
...
+		phoneNumberValidation.checkPhoneNumberValidation("902 855 3461");
+		phoneNumberValidation.checkPhoneNumberValidation("9028 55 3461");
+		phoneNumberValidation.checkPhoneNumberValidation("90i 855 3461");
+		phoneNumberValidation.checkPhoneNumberValidation("202 855 3461");
+		System.out.println("\n" +"phone number of length :14");
+		phoneNumberValidation.checkPhoneNumberValidation("+91 9028553461");
+		phoneNumberValidation.checkPhoneNumberValidation("+91 90 8553461");
+		phoneNumberValidation.checkPhoneNumberValidation("+91 90y8553461");
+		phoneNumberValidation.checkPhoneNumberValidation("+91 9028553461");
+		phoneNumberValidation.checkPhoneNumberValidation("+91 2028553461");
```
AavrutiDJun20: Instead of calling same method multiple times you can create an array of String and call the method in the loop and there is no need to print the length of the Mobile Number.

ArchanaNJun20: Hi Aavruti, The code has updated as per your given comment.

AavrutiDJun20 (approved):
Code looks good to me

harshvegada (approved):
Code Looks good,
its working as expected.

PR #748, opened 2020-07-07
Harshad assignment 23

HarshadBJun20:
-> Assignment 23 Done.
-> Program 1 Screenshot of output:
image

-> Program 2 Screenshot of output:
image

Please review.

### DIFF
--- a/src/harshadBJun20/Assignment24/ValidateMobileNumber.java
+++ b/src/harshadBJun20/Assignment24/ValidateMobileNumber.java
@@ -1,0 +1,60 @@
+package harshadBJun20.Assignment24;
+
+/*Assignment 24 :- 5th Jul'2020 
+Validate given phone number.
+
++91 9765463742 --> valid
+09765463742 --> valid
+976 546 3742 --> valid
+9765463742 --> valid
+
+9765463A42 --> Invalid
+976546342 --> Invalid
+6976546342 --> Invalid
+
+Note : phone number is given in String format.*/
+
+public class ValidateMobileNumber {
+
+	void checkMobileNumberValidOrNot(String mobNumber){
+		String checkMobNumber = mobNumber;
+		boolean flag=false;
+		if(mobNumber.length()>=10) {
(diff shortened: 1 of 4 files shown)


src/harshadBJun20/Assignment24/ValidateMobileNumber.java
```diff
@@ -0,0 +1,60 @@
...
+		String checkMobNumber = mobNumber;
+		boolean flag=false;
+		if(mobNumber.length()>=10) {
+			if(mobNumber.startsWith("+91") || mobNumber.startsWith("0",0) || mobNumber.startsWith("9",0) || mobNumber.startsWith("8",0) || mobNumber.startsWith("7",0)) {
+				flag = true;
+				mobNumber = mobNumber.replace("+","").replace(" ","");
+
+				for(int index=0;index<mobNumber.length();index++) {
+					if(Character.isDigit(mobNumber.charAt(index))) {
+						flag = true;
```
KrishnaKTechnocredits, line 29: If and else can be replaced with only if.
if(!Character.isDigit(mobNumber.charAt(index))) {
						flag = false;
						break;
					}

src/harshadBJun20/Assignment24/ValidateMobileNumber.java
```diff
@@ -0,0 +1,60 @@
...
+
+		if(flag == true) {
+			System.out.println(checkMobNumber+" --> Valid");
+		}else {
+			System.out.println(checkMobNumber+" --> Invalid");
+		}
+	}
+	public static void main(String[] args) {
+		ValidateMobileNumber validateMobileNumber = new ValidateMobileNumber();
+		validateMobileNumber.checkMobileNumberValidOrNot("+91 9765463742");
```
KrishnaKTechnocredits, line 52: +91 2765463742
Program shows number is valid for above number. It should be invalid as number should start with 7,8,9 after +91.

Even it should not allow, +91 2765 46 3742. extra 2 spaces are there in phone number

PR #749, opened 2020-07-08
AMITA_Assignment_25_Done_10mins

AmitaRJun20:
AMITA_Assignment_25_Done...Please Review...
SumofnumbersinString

### DIFF
--- a/src/amitaRJun20/SumOfAllNumbersInString.java
+++ b/src/amitaRJun20/SumOfAllNumbersInString.java
@@ -5,26 +5,24 @@ Input: J7yu9y8h1h8j4b7j3jjb6
 Ouput: Sum of all numbers in String is: 54*/
 package amitaRJun20;
 
-import java.util.Arrays;
-
 public class SumOfAllNumbersInString {
 	
 	int sum = 0;
 	
 	//method to display sum of all numbers in input String
-	void displaySum(String input) {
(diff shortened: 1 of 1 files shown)


src/amitaRJun20/SumOfAllNumbersInString.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class SumOfAllNumbersInString {
+	
+	int sum = 0;
+	
+	//method to display sum of all numbers in input String
+	void displaySum(String input) {
+		for (int index = 0; index < input.length(); index++) {
+			if (Character.isDigit(input.charAt(index))) {
+				int number = Integer.parseInt(String.valueOf(input.charAt(index)));
```
MaheshKJun20: We can avoid creating number variable and directly add this value using the expression

MaheshKJun20 (approved):
Hi Amit, 
Code looks good. Approving it for further review.
Have added a suggestion at one place you can check on that

src/amitaRJun20/SumOfAllNumbersInString.java
AmitaRJun20: updated as per suggestion...Thank you for feedback..

KrishnaKTechnocredits (approved):
it looks good to me.

PR #758, opened 2020-07-08
Prashant_Assignment_25_30mins

ghost:
Assignment for WAP to sum all the numbers in given string Done.
Assignment_25

### DIFF
--- a/src/prashantCJune20/SumofIntinString.java
+++ b/src/prashantCJune20/SumofIntinString.java
@@ -1,0 +1,19 @@
+package prashantCJune20;
+
+public class SumofIntinString {
+	
+	void addDigits(String input) {
+		int addition = 0;
+		for(int index = 0;index<input.length();index++) {
+			if(Character.isDigit(input.charAt(index)))
+				addition = addition	+ Integer.parseInt(String.valueOf(input.charAt(index)));
+		}
+		System.out.println("Addition of your numbers between the characters is  "+addition);		
+	}
+		public static void main(String[] args) {		
+		SumofIntinString sum = new SumofIntinString();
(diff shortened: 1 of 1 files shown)


src/prashantCJune20/SumofIntinString.java
```diff
@@ -0,0 +1,19 @@
+package prashantCJune20;
+
+public class SumofIntinString {
+	
+	void addDigits(String input) {
+		int addition = 0;
+		for(int index = 0;index<input.length();index++) {
+			if(Character.isDigit(input.charAt(index)))
+				addition = addition	+ Integer.parseInt(String.valueOf(input.charAt(index)));
```
ashishpawar517: Inside the If block instead this long statement 
```java 
addition = addition	+ Integer.parseInt(String.valueOf(input.charAt(index)));
```

You could simply do 
```java 
addition += input.charAt(index)-'0';
```

src/prashantCJune20/SumofIntinString.java
```diff
@@ -0,0 +1,19 @@
+package prashantCJune20;
+
+public class SumofIntinString {
+	
+	void addDigits(String input) {
+		int addition = 0;
+		for(int index = 0;index<input.length();index++) {
```
ashishpawar517: index of the  character isn't necessary for the execution you can simply just use `forEach` loop

src/prashantCJune20/SumofIntinString.java
ghost: Logic changed as per comments.

KrishnaKTechnocredits (approved):
Code looks good to me.

ashishpawar517 (commented):
done !

PR #770, opened 2020-07-08
Abhishek_Assignment_25_30minute

AbhishekSJune2020:
output:
Sum of all numbers in String is : 53

### DIFF
--- a/src/abhishekSJun20/SumOfNumbersInString.java
+++ b/src/abhishekSJun20/SumOfNumbersInString.java
@@ -1,0 +1,25 @@
+package abhishekSJun20;
+/*WAP to sum all the numbers in given string.
+	Input: J7yu9y8h1h8j4b7j3jjb6
+	Output: Sum of all numbers in String is: 53
+ */
+
+public class SumOfNumbersInString {
+//
+	void sumOfNumber(String input) {
+		int sum = 0;
+		for (int index = 0; index < input.length(); index++) {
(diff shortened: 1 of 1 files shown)


PiyushPJun20:
Code looks good to me.
one suggestion, Instead of storing in Number variable, we can optimize code by using:
sum + = Integer.parseInt(String.valueOf(input.charAt(index)));

PiyushPJun20 (commented):
Code looks good to me.
one suggestion, Instead of storing in Number variable, we can optimize code by using:
sum + = Integer.parseInt(String.valueOf(input.charAt(index)));

src/abhishekSJun20/SumOfNumbersInString.java
```diff
@@ -0,0 +1,24 @@
...
+	Output: Sum of all numbers in String is: 53
+ */
+
+public class SumOfNumbersInString {
+
+	void sumOfNumber(String input) {
+		int sum = 0;
+		for (int index = 0; index < input.length(); index++) {
+			if (Character.isDigit(input.charAt(index))) {
+				sum += Integer.parseInt(String.valueOf(input.charAt(index)));
```
KrishnaKTechnocredits, line 13: You can use below method to convert char to int
Character.getNumericValue(numberString.charAt(index));

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #775, opened 2020-07-08
YashC_Assignment 25_Done_20mins"

YashCJun20:
image

### DIFF
--- a/src/yashCJun20/Assignment25/SumOfAllNumbersInString.java
+++ b/src/yashCJun20/Assignment25/SumOfAllNumbersInString.java
@@ -2,7 +2,7 @@ package yashCJun20.Assignment25;
 
 public class SumOfAllNumbersInString {
 	
-	void sum (String name) {
+	void toCalculateSumOfAllNumbersInString (String name) {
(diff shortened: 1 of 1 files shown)


src/yashCJun20/Assignment25/SumOfAllNumbersInString.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class SumOfAllNumbersInString {
+	
+	void toCalculateSumOfAllNumbersInString (String name) {
+		
+			int sum=0;
+			for(int i=0;i<name.length();i++) {
+				if(Character.isDigit((name.charAt(i)))) {
+					String charValue = String.valueOf(name.charAt(i));
+					int digit = Integer.parseInt(charValue);
```
KrishnaKTechnocredits, line 11: Instead of line number 10 and 11. you can write below line.

You can use below method to convert char to int
Character.getNumericValue(numberString.charAt(index));

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #777, opened 2020-07-08
Kartikey assignment 24

KartikeyDJune20:
image

### DIFF
--- a/src/kartikeyDJune20/PhoneNumValidations.java
+++ b/src/kartikeyDJune20/PhoneNumValidations.java
@@ -1,0 +1,40 @@
+package kartikeyDJune20;
+
+public class PhoneNumValidations {
+	
+		
+		void phoneNumValidation(String number) {
+			int numLength=number.length();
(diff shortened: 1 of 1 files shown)


src/kartikeyDJune20/PhoneNumValidations.java
```diff
@@ -0,0 +1,35 @@
+package kartikeyDJune20;
+
+public class PhoneNumValidations {		
+		void phoneNumValidation(String number) {
+			int numLength=number.length();
+			boolean isNumValid=false;
+			if(numLength>=10) {
+				for(int index=0;index<number.length();index++) {			
+					if(numLength==11 && number.charAt(0)=='0') {
```
harshvegada, line 9: here making flag value as true, not proper logic, like If in between once the flag value set as false but again loop iterating and for next value flag comes true but if once the flag value is false then you need to break n say that number is invalid.

put if condition in loop to check flag value

PR #779, opened 2020-07-08
Anup_Assignment_25_Done_10mins_PR

AnupSJun20:
Please review.
Program1

### DIFF
--- a/src/anupSJun20/AddIntegers.java
+++ b/src/anupSJun20/AddIntegers.java
@@ -1,0 +1,22 @@
+package anupSJun20;
+/*WAP to sum all the numbers in given string.
+Input: J7yu9y8h1h8j4b7j3jjb6
+Ouput: Sum of all numbers in String is: 53*/
+
+public class AddIntegers {
+	int sum = 0;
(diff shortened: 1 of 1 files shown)


src/anupSJun20/AddIntegers.java
```diff
@@ -0,0 +1,22 @@
...
+/*WAP to sum all the numbers in given string.
+Input: J7yu9y8h1h8j4b7j3jjb6
+Ouput: Sum of all numbers in String is: 53*/
+
+public class AddIntegers {
+	int sum = 0;
+	void findSumOfIntegers(String str) {
+		for(int index = 0;index<str.length();index++) {
+			if(Character.isDigit(str.charAt(index))) {
+				sum += Integer.parseInt(String.valueOf(str.charAt(index)));
```
KrishnaKTechnocredits, line 11: Here you should use Character class method to convert char to int.
Character.getNumericValue(ch).
Better optimized way.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #784, opened 2020-07-09
ADT_Assignment26_25Mins

AditiGJUN20:
### Assignment 26:
**Time:** 25 Mins
For the code i was little confused about what Maulik had mentioned about last condition check. I've not added any condition but i'm getting output for various String inputs as expected.
Let me know if code needs any correction or further improvements.
Screen Shot 2020-07-09 at 12 43 36 AM

### DIFF
--- a/src/aditiGJUN20/SumOfDigits.java
+++ b/src/aditiGJUN20/SumOfDigits.java
@@ -1,0 +1,42 @@
+// Find sum of all digits in given string. (Consider consecutive digits till next character as single number)
+
+package aditiGJUN20;
+
+public class SumOfDigits {
+	int getSumofAllDigits(String str) {
+		// Condition for empty String
+		if (str.isEmpty()) {
+			System.out.println("Null String");
+			return 0;
+		}
+		
+		int digitSum = 0;
+		String temp = "0"; // Earlier tried with "" and null but had thrown exception so settled with 0 as
+							// doing integer sum.
+		for (int index = 0; index < str.length(); index++) {
+			char ch = str.charAt(index);
+			if (Character.isDigit(ch))
+				temp += ch;
+			else {
+				digitSum += Integer.parseInt(temp);
+				temp = "0";
+			}
+		}
+		int finalSum = digitSum + Integer.parseInt(temp);
+		if (finalSum == 0 && str.contains("0"))
+			System.out.println("Input String contains only '0' as digit");
+		return finalSum;
+	}
+
+	public static void main(String[] args) {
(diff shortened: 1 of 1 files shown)


src/aditiGJUN20/SumOfDigits.java
```diff
@@ -0,0 +1,42 @@
...
+public class SumOfDigits {
+	int getSumofAllDigits(String str) {
+		// Condition for empty String
+		if (str.isEmpty()) {
+			System.out.println("Null String");
+			return 0;
+		}
+		
+		int digitSum = 0;
+		String temp = "0"; // Earlier tried with "" and null but had thrown exception so settled with 0 as
```
AmitaRJun20: I can think of one approach instead of declaring String temp = "0", you can declare String temp ="" and handle the condition in else part which will avoid number format exception..
else {
				if(temp.length() !=0) {
					sum += Integer.parseInt(temp);
					temp = "";
				}

src/aditiGJUN20/SumOfDigits.java
```diff
@@ -0,0 +1,42 @@
...
+			char ch = str.charAt(index);
+			if (Character.isDigit(ch))
+				temp += ch;
+			else {
+				digitSum += Integer.parseInt(temp);
+				temp = "0";
+			}
+		}
+		int finalSum = digitSum + Integer.parseInt(temp);
+		if (finalSum == 0 && str.contains("0"))
```
AmitaRJun20: Here we can avoid creating another finalSum variable..
 if the last digit is a number, then again one validation is required outside of the For loop:

if(temp.length() != 0) {
			sum += Integer.parseInt(temp);
		}
		return sum;

AmitaRJun20 (commented):
I have added some suggestion ...Please go through that...

src/aditiGJUN20/SumOfDigits.java
AditiGJUN20: Thank you @AmitaRJun20 for your insights. I've added extra condition and handled the code as you suggested. Now the code is giving correct output with `temp=""`

src/aditiGJUN20/SumOfDigits.java
AditiGJUN20: @AmitaRJun20 
I was referring to this condition only which i'd missed. I've added it in code and removed the extra variable.
Thank you for detailed review.

AmitaRJun20 (approved):
Code looks good ...Approved...

KrishnaKTechnocredits (approved):
@AditiGJUN20 this is as expected. Good job.

PR #788, opened 2020-07-09
Anshu-Assignment-26-1hr-Done

AnshuKJun20:
image


PranitaPJun20 (approved):
The output is exactly as expected.
You can write the code by initializing 2 local variables also.
Try to write the code the way Maulik explained.That will be easy I think.

KrishnaKTechnocredits (approved):
Anshu code looks good to me but it could be written in more simpler way. Pranita is correct. Take a look at https://github.com/KrishnaKTechnocredits/JAVATechnoJun20/pull/789

PR #791, opened 2020-07-09
Ajit_Assignment_26_Done_25mins

AjitSJune20:
image

### DIFF
--- a/src/ajitSJune20/SumOfNumInString.java
+++ b/src/ajitSJune20/SumOfNumInString.java
@@ -1,0 +1,28 @@
+package ajitSJune20;
+
+public class SumOfNumInString {
+
+	public static int getSumOfAllDigits(String input) {
+		int sum = 0;
+		String temp = "";
+		for (int index = 0; index < input.length(); index++) {
+			char ch = input.charAt(index);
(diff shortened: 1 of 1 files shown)


src/ajitSJune20/SumOfNumInString.java
```diff
@@ -0,0 +1,28 @@
...
+		int sum = 0;
+		String temp = "";
+		for (int index = 0; index < input.length(); index++) {
+			char ch = input.charAt(index);
+			if (Character.isDigit(ch))
+				temp += ch;
+			else {
+				if (temp.length() != 0) {
+					sum += Integer.parseInt(temp);
+					temp = "0";
```
KartikeyDJune20, line 15: If you are reinializing temp with "0".. that means your length will never gonna be zero again, so there is no point for the if statement above to check the length

KartikeyDJune20: Oh got it for the first time it will need that..

AjitSJune20: Thanks.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #796, opened 2020-07-09
AMITA_Assignment_26_Done_20mins

AmitaRJun20:
Please Review Assignment_26..Thank you..
SumOfIntergersInString

### DIFF
--- a/src/amitaRJun20/SumOfConsecutiveIntegersInString.java
+++ b/src/amitaRJun20/SumOfConsecutiveIntegersInString.java
@@ -17,7 +17,7 @@ public class SumOfConsecutiveIntegersInString {
 		String temp = "";
 		for (int index = 0; index < input.length(); index++) {
 			if (Character.isDigit(input.charAt(index))) {
-				temp += String.valueOf(input.charAt(index));
+				temp += (input.charAt(index));
 			} else {
 				if (temp.length() != 0) {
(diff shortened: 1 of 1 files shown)


src/amitaRJun20/SumOfConsecutiveIntegersInString.java
```diff
@@ -0,0 +1,41 @@
...
+public class SumOfConsecutiveIntegersInString {
+	
+	//method to find sum of all integers in a String
+	int findSumOfIntegers(String input) {
+		System.out.println("\nThe given String is " + input);
+		int sum = 0;
+		String temp = "";
+		for (int index = 0; index < input.length(); index++) {
+			if (Character.isDigit(input.charAt(index))) {
+				temp += String.valueOf(input.charAt(index));
```
MaheshKJun20: String.Valueof method is not required here. we can add String + char which will return String.

MaheshKJun20 (approved):
Hi Amita,
Reviewed your PR. Code is working as expected. 
Recommended a small change. 
Approving the PR.

src/amitaRJun20/SumOfConsecutiveIntegersInString.java
AmitaRJun20: You are right Mahesh...thanks for the feedback..updated the code...

PR #802, opened 2020-07-09
Suparna assignment 26_10_min 

SuparnaNikam:
SumOfNo
Stringnosum

### DIFF
--- a/src/suparnaNJune20/Assignment_26/SumOFNoFromStr.java
+++ b/src/suparnaNJune20/Assignment_26/SumOFNoFromStr.java
@@ -22,6 +22,8 @@ public class SumOFNoFromStr {
 					} else {
 						sum = sum + tempNo;
 						break;
+						
+						
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


src/suparnaNJune20/Assignment_26/SumOFNoFromStr.java
```diff
@@ -0,0 +1,43 @@
...
+			int sum = 0, tempNo = 0;
+		for (int outerIndex = 0; outerIndex < str.length(); outerIndex++) {
+			tempNo = 0;
+			if (Character.isDigit(str.charAt(outerIndex))) {
+				for (; outerIndex < str.length(); outerIndex++) {
+					if (Character.isDigit(str.charAt(outerIndex))) {
+						tempNo = tempNo * 10 + Character.getNumericValue(str.charAt(outerIndex));
+					} else {
+						sum = sum + tempNo;
+						break;
```
VireshJJun20, line 24: Before break, you need to reset the tempNo to 0. and before printing the sum, need to check the tempNo value. 
try with this input: P171Rr29T2E >> output should be 202 but your logic returns 204

VireshJJun20 (changes requested):
Please update the logic as per suggestion.

PR #815, opened 2020-07-12
Viresh_TreeProgram_8mins

VireshJJun20:
Please review and comment.

image

### DIFF
--- a/src/TreeProgram.java
+++ b/src/TreeProgram.java
@@ -1,12 +1,12 @@
 
 public class TreeProgram {
 	
-	int Max=10, count=1;
+	// removed instance variables.
 	void findVisibleTree(int[] arr){
-		int Max=arr[0], count=1;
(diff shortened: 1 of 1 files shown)


src/TreeProgram.java
```diff
@@ -0,0 +1,21 @@
+
+public class TreeProgram {
+	
+	// removed instance variables.
+	void findVisibleTree(int[] arr){
+		int max=arr[0], count=1; // updated variable name as max.
+		for (int index= 0; index < arr.length; index++) {
```
harshvegada, line 7: Index should start from 1 as we already define max variable value from 0 so, we compare value from index 1
you know the changes already

harshvegada (approved):
Merging PR
Just logical change

src/TreeProgram.java
VireshJJun20: Thanks Harsh, yes I got this point and completely agree on this.

PR #816, opened 2020-07-12
Palak_TreeProgram_8Min

PalakSJun20:
image


PalakSJun20:
image

harshvegada (approved):
Index should starts   from 1, if you already initialised max variable value from 0
just one logical changes and count value should starts from 1 as we can see first tree in any condition

PR #822, opened 2020-07-12
AMITA_TreeProgram_Done_12mins

AmitaRJun20:
AMITA_TreeProgram_Done_12mins

### DIFF
--- a/src/amitaRJun20/TreeProgram.java
+++ b/src/amitaRJun20/TreeProgram.java
@@ -4,21 +4,20 @@ public class TreeProgram {
 	
 	void display(int[] input) {
 		
-		int count =0;
+		int count =1;
+		int max =input[0];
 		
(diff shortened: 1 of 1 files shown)


src/amitaRJun20/TreeProgram.java
```diff
@@ -0,0 +1,24 @@
+package amitaRJun20;
+
+public class TreeProgram {
+	
+	void display(int[] input) {
+		
+		int count =0;
+		
+		for(int i =0; i<input.length-1; i++) {
+			int max =input[i];
```
harshvegada: this line should outer side of for loop,

> Index should   starts from 1, if you already initialised max variable value from a[0], and   
> counter value should start from 1 as we can see the first tree any condition

harshvegada (approved):

PR #824, opened 2020-07-12
Madhura_TreeProgram_18mins

MadhuraMJun20:


src/madhuraMJun20/TreeProgram.java
```diff
@@ -0,0 +1,25 @@
+package madhuraMJun20;
+//[3,5,5,11,9,10,13]
+public class TreeProgram {
+	void findNumOfTrees(int [] arr) {
+		int max = 0;
+		int cnt=1;
+		for(int index = 1; index < arr.length-1; index++) {
```
harshvegada: this logic looks like for second max number
Code should be like below
image

PR #841, opened 2020-07-14
Shruti Assignment 27- 20 mints

ShrutiDJun20:
image

### DIFF
--- a/src/shrutiDJun20Assignments/MeetingRoom.java
+++ b/src/shrutiDJun20Assignments/MeetingRoom.java
@@ -3,13 +3,13 @@ package shrutiDJun20Assignments;
 public class MeetingRoom {
 
 	String room1 = "Nalanda";
-	static int cnt1 = 0;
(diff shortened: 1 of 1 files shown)


PalakSJun20:
As in requirement nothing is mentioned regarding maximum hours limit . 
Method Overloaded method is used , Code Looks good to me 
Krishna can u check once

PalakSJun20 (approved):

src/shrutiDJun20Assignments/MeetingRoom.java
```diff
@@ -0,0 +1,114 @@
+package shrutiDJun20Assignments;
+
+public class MeetingRoom {
+
+	boolean nalandaBookingStatus, takshashilaBookingStatus,
```
KrishnaKTechnocredits, line 5: Here this all variable should be static only, what If I will create new object of class and check. the. status.

PR #843, opened 2020-07-14
Pranita_Assignment_27_Done_30mins

PranitaPJun20:
image

### DIFF
--- a/src/pranitaPJun20/MeetingRoomOverloading.java
+++ b/src/pranitaPJun20/MeetingRoomOverloading.java
@@ -1,0 +1,67 @@
+package pranitaPJun20;
+
+public class MeetingRoomOverloading {
+	int bookingTime;
+	String selectedMeetingRoom;
+	String[] rooms = { "Nalanda", "Takshashila", "AgraFort", "PratapGadh" };
+
+	void bookMeetingRoom() {
+		selectedMeetingRoom = "Nalanda";
+		bookingTime = 1;
+		displayInfo();
+		for (int index = 0; index < rooms.length; index++) {
+			if (rooms[index].equals("Nalanda")) {
+				rooms[index] = " ";
+			}
+		}
+	}
+
+	void bookMeetingRoom(String meetingRoomName) {
+		if (!meetingRoomName.equals("Nalanda")) {
+			selectedMeetingRoom = meetingRoomName;
+			bookingTime = 1;
+			for (int index = 0; index < rooms.length; index++) {
+				if (rooms[index].equals(meetingRoomName)) {
+					rooms[index] = " ";
+				}
+			}
+			displayInfo();
+		} else {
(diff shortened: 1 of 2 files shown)


src/pranitaPJun20/MeetingRoomOverloading.java
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+	}
+
+	void bookMeetingRoom(String meetingRoomName, int time) {
+		selectedMeetingRoom = meetingRoomName;
+		bookingTime = time;
+		for (int index = 0; index < rooms.length; index++) {
+			if (rooms[index].equals(meetingRoomName)) {
+				rooms[index] = " ";
```
harshvegada, line 40: Hi @PranitaPJun20 ,
Here once you entered in for loop **break** it because no need to iterate for other rooms,

src/pranitaPJun20/MeetingRoomOverloading.java
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+	}
+
+	public static void main(String[] args) {
+		MeetingRoomOverloading meetingRoomOverloading = new MeetingRoomOverloading();
+		meetingRoomOverloading.bookMeetingRoom();
+		meetingRoomOverloading.bookMeetingRoom("Nalanda");
+		meetingRoomOverloading.bookMeetingRoom("Takshashila");
+		meetingRoomOverloading.bookMeetingRoom("AgraFort", 1);
```
harshvegada, line 63: here you call displayInfo method in every method, i want that method call by customer to check which is booked for how many hours, for that check below hint:

what you should do is define String array named as meetingRoomName (already in your code) and second array which contains hours.

in displaInfor() you will use for loop and check, if hours is 0 then on same index room is also available.

int hours[] = {0,0,0,0};

before book any room, in your code you should add method named as isRoomAvailable() [this method checks on hours array if any slot is available than return true to book room]

harshvegada (commented):
work on comment

src/pranitaPJun20/MeetingRoomOverloading.java
PranitaPJun20: Pushed new file with proper changes.

src/pranitaPJun20/MeetingRoomOverloading.java
PranitaPJun20: Thank you for the detailed explanation.I uploaded new files with all code changes.Please review that.

harshvegada: Okay @PranitaPJun20

harshvegada (approved):
very good @PranitaPJun20 
No comments, as per the expectation

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #844, opened 2020-07-14
Vaishnavi_Assignment_27_35Min_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20_arrayExamples/MeetingRoom.java
+++ b/src/vaishnaviVJun20_arrayExamples/MeetingRoom.java
@@ -1,72 +1,80 @@
 package vaishnaviVJun20_arrayExamples;
 
+/*
+Program to demonstrate overloading behavior.
+Create a class MeetingRoom having method bookMeetingRoom which should provide
+below facility:
+meeting room names: Nalanda, Takshashila, AgraFort, and PratapGadh
+a. If any user call method bookMeetingRoom() at that time“Nalanda” meeting room will be booked for 1 hr.
+b. Write a method that helps to book a meeting room other than "Nalanda" for default timing would be 1 hr.
+c. Write a method which helps to book a meeting room by giving the name of the meeting room and given hrs.
+d. displayInfo() method should print booked meeting room info [i.e. meeting room,
+booked for how many hrs].
+e. displayAvilableMeetingRoomForBooking() method will print available meeting room names.
+[i.e. If Nalanda and Takshashila is booked then this method should print "Available meeting rooms: AgraFort and
+PratapGadh"]
+*/
+
 import java.util.Scanner;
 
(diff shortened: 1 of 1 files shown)


src/vaishnaviVJun20_arrayExamples/MeetingRoom.java
```diff
@@ -0,0 +1,93 @@
...
+		meetingRoom.bookMeetingRoom();
+		System.out.println("Which Meeting room would you like to book : Takshashila/AgraFort/PratapGadh ? ");
+		String meetingRoomName = scanner.next();
+		meetingRoom.bookMeetingRoom(meetingRoomName);
+		System.out.println("Which Meeting room would you like to book and for how many hours : Takshashila/AgraFort/PratapGadh/Nalanda ? ");
+		meetingRoomName = scanner.next();
+		int hrs = scanner.nextInt();
+		meetingRoom.bookMeetingRoom(meetingRoomName, hrs);
+		meetingRoom.displayAvilableMeetingRoomForBooking();
+		scanner.close();
```
harshvegada: Hi @VaishnaviVJun20,
Here while calling method you are printing time for booked room but you need to check which for room booked for how many hours, this FR not fulfilled in code

what you should do is define String array named as meetingRoomName and second array which contains hours.

int hours[] = {0,0,0,0};

harshvegada (commented):
work on comment

src/vaishnaviVJun20_arrayExamples/MeetingRoom.java
VaishnaviVJun20: -- Updated the code, Please review

src/vaishnaviVJun20_arrayExamples/MeetingRoom.java
```diff
@@ -0,0 +1,122 @@
...
+				takshashilaHours = hours;
+			} else
+				System.out.println(meetingRoomName + " meeting room is already booked for " + hours + "hr.");
+		} else if (meetingRoomName == MEETINGROOMS.AgraFort) {
+			if (bookingStatusForAgraFort == false) {
+				System.out.println(meetingRoomName + " meeting room is booked for " + hours + "hr.");
+				bookingStatusForAgraFort = true;
+				agraFortHours = hours;
+			} else
+				System.out.println(meetingRoomName + " meeting room is already booked for " + hours + "hr.");
```
harshvegada, line 62: Hi @VaishnaviVJun20 
if you check this line gives wrong information, line 48, 55 and 62
this gives wrong informart

harshvegada (approved):
looks rest thing good work in

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #852, opened 2020-07-14
Archana_Assisgment_27_45mins_Done

ArchanaNJun20:
Updated Enum
image

### DIFF
--- a/src/archanaNJun20/BookMeetingRoom.java
+++ b/src/archanaNJun20/BookMeetingRoom.java
@@ -1,0 +1,105 @@
+package archanaNJun20;
+
+public class BookMeetingRoom {
+
+	String meetingRoom;
+	double hours=0.0;
+	boolean nalandaFlag=false;
+	boolean takshashilaFlag=false;
+	boolean agraFortFlag=false;
+	boolean pratapGadhFlag=false;
+
+	void bookMeetingRoom(){
+		if(nalandaFlag==false){
+			meetingRoom="Nalanda";
+			hours=1;
+			displayInfo();
+			nalandaFlag=true;
+		}
+		else{
+			System.out.println("-->The meeting room Nalanda is already booked.");
+		}
+	}
+	void bookMeetingRoom(String meetingRoom){
+		if(meetingRoom!="Nalanda"){
+			if(meetingRoom=="Takshashila"&&takshashilaFlag==false){
+				this.meetingRoom="Takshashila";
+				hours=1;
+				displayInfo();
+				takshashilaFlag=true;
+			}
+			else if(meetingRoom=="AgraFort" &&agraFortFlag==false){
+				this.meetingRoom="AgraFort";
+				hours=1;
+				displayInfo();
+				agraFortFlag=true;
+			}
+			else if(meetingRoom=="PratapGadh"&&pratapGadhFlag==false){
+				this.meetingRoom="PratapGadh";
+				hours=1;
+				displayInfo();
+				pratapGadhFlag=true;
+			}
+			else{
+				System.out.println(meetingRoom +"-->The meeting room Nalanda is already booked.");
+			}
+		}
+		else{
+			System.out.println(meetingRoom +"-->The meeting room Nalanda is already booked.");
+		}
+	}
+	void bookMeetingRoom(String meetingRoom,int hours){
+		if(meetingRoom!="Nalanda"){
+			if(meetingRoom=="Takshashila"&&takshashilaFlag==false){
+				this.meetingRoom="Takshashila";
+				this.hours=hours;
+				displayInfo();
+				takshashilaFlag=true;
+			}
+			else if(meetingRoom=="AgraFort"&&agraFortFlag==false){
+				this.meetingRoom="AgraFort";
+				this.hours=hours;
+				displayInfo();
+				agraFortFlag=true;
+			}
+			else if(meetingRoom=="PratapGadh"&&pratapGadhFlag==false){
+				this.meetingRoom="PratapGadh";
+				this.hours=hours;
+				displayInfo();
+				pratapGadhFlag=true;
+			}else{
+				System.out.println(meetingRoom+ "-->The meeting room  is already booked.");
+			}
+		}
+		else{
+			System.out.println("-->The meeting room Nalanda is already booked.");
+		}
+	}
+	void displayInfo(){
+		System.out.println( meetingRoom+" Meeting room "  + " is booked for " + hours + " hour");
+	}
+	void displayAvilableMeetingRoomForBooking() {
+		System.out.println("\nAvailable meeting Rooms:");
+		if(pratapGadhFlag==false){
+			System.out.println("*PratapGadh metting room is available*");
+		}
+		if(agraFortFlag==false)	{
+			System.out.println("*AgraFort metting room is available*");
+		}
+		if(nalandaFlag==false){
+			System.out.println("*Nalanda metting room is available*");
+		}
+		if(takshashilaFlag==false){
+			System.out.println("*Takshashila metting room is available*");
+		}
+	}
+	public static void main(String[] args) {
+		MeetingRoom meetingRoom=new MeetingRoom();
+		meetingRoom.bookMeetingRoom();
+		meetingRoom.bookMeetingRoom();
+		meetingRoom.bookMeetingRoom("AgraFort");
+		meetingRoom.bookMeetingRoom("AgraFort",1);
+		meetingRoom.bookMeetingRoom("Takshashila",2);
+		meetingRoom.displayAvilableMeetingRoomForBooking();
+	}
+}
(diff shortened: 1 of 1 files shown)


src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+			hours=1;
+			displayInfo();
+			nalandaFlag=true;
+		}
+		else{
+			System.out.println("-->The meeting room Nalanda is already booked.");
+		}
+	}
+	void bookMeetingRoom(String meetingRoom){
+		if(meetingRoom!="Nalanda"){
```
AavrutiDJun20: For String variable comparison we should write like this '!meetingRoom.equals("Nalanda")'

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+			if(meetingRoom=="Takshashila"&&takshashilaFlag==false){
```
AavrutiDJun20: For String variable comparison we should write like this 'meetingRoom.equals("Takshashila")'

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+				agraFortFlag=true;
+			}
+			else if(meetingRoom=="PratapGadh"&&pratapGadhFlag==false){
+				this.meetingRoom="PratapGadh";
+				hours=1;
+				displayInfo();
+				pratapGadhFlag=true;
+			}
+			else{
+				System.out.println(meetingRoom +"-->The meeting room Nalanda is already booked.");
```
AavrutiDJun20: Here SOP should print "Meeting room is already booked"

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+				displayInfo();
+				agraFortFlag=true;
+			}
+			else if(meetingRoom=="PratapGadh"&&pratapGadhFlag==false){
+				this.meetingRoom="PratapGadh";
+				this.hours=hours;
+				displayInfo();
+				pratapGadhFlag=true;
+			}else{
+				System.out.println(meetingRoom+ "-->The meeting room  is already booked.");
```
AavrutiDJun20, line 71: Here SOP should print "Invalid meeting room name" as we are not going to go inside if condition if Meeting Room name is "Nalanda"

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+		}
+		if(agraFortFlag==false)	{
+			System.out.println("*AgraFort metting room is available*");
+		}
+		if(nalandaFlag==false){
+			System.out.println("*Nalanda metting room is available*");
+		}
+		if(takshashilaFlag==false){
+			System.out.println("*Takshashila metting room is available*");
+		}
```
AavrutiDJun20, line 72: in this you can add condition like what if all the meeting rooms are booked

src/archanaNJun20/BookMeetingRoom.java
ArchanaNJun20: The code is updated as per your suggestion. Thank you.

src/archanaNJun20/BookMeetingRoom.java
ArchanaNJun20: The code is updated as per your suggestion. Thank you.

src/archanaNJun20/BookMeetingRoom.java
ArchanaNJun20: The code is updated as per your suggestion. Thank you.

src/archanaNJun20/BookMeetingRoom.java
ArchanaNJun20: The code is updated as per your suggestion. Thank you.

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,112 @@
...
+				hours=1;
+				displayInfo();
+				pratapGadhFlag=true;
+			}
+			else{
+				System.out.println(meetingRoom +"-->The meeting room is already booked.");
+			}
+		}
+		else{
+			System.out.println(meetingRoom +" is Invalid meeting Room ");
```
AavrutiDJun20: This statement will show "Nalanda is invalid meeting room" if user passes "Nalanda" as a value.
for Invalid meeting room you will have to add one more if condition to check all the valid meeting room names

ArchanaNJun20: The code is updated as per your suggestion.

AavrutiDJun20: it will still show "Nalanda is invalid meeting room"
you can write like this:

if(meetingRoom.equals("Nalanda"){
SOP("Nalanda is already booked"); Or you can call BookMeeetingRoom() method from here
}
else if(meetingRoom.equals("Takshashila") || meetingRoom.equals("AgraFort") || meetingRoom.equals("PratapGadh")){

}
else{
SOP("Invalid Room");
}

Also update the Output screenshot

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+		if(meetingRoom==MEETINGROOM.Nalanda){
+			if(nalandaFlag==false){
+				System.out.println(meetingRoom+" meeting room is booked for "+hours+"hr");
+				nalandaHours=hours;
+				nalandaFlag=true;
+			}
+			else 
+				System.out.println("Nalanda meeting room  is already booked.");
+		}else if(meetingRoom==MEETINGROOM.Takshashila) {
+			System.out.println(meetingRoom+" meeting room is booked for "+hours+"hr");
```
AavrutiDJun20, line 22: this line should be inside if(takshashilaFlag==false)

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+		if(takshashilaFlag){
+			System.out.println("Takshashila metting room is Booked for "+takshashilaHours+"hr");
+		}
+	}
+	void displayAvilableMeetingRoomForBooking() {
+		System.out.println("\nAvailable Meeting Rooms:");
+		if(!pratapGadhFlag){
+			System.out.println("PratapGadh metting room is available");
+		}
+		if(!agraFortFlag)	{
```
AavrutiDJun20: remove space between brackets

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
+package archanaNJun20;
+
+public class BookMeetingRoom {
+	
+	double nalandaHours,takshashilaHours,agraFortHours,pratapGadhHours;
```
AavrutiDJun20, line 4: No need to define hours variables of double type as you have used "int hours" in whole program while passing it as a parameter

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+			System.out.println("Nalanda meeting  room  is already booked.");
+		}
+	}
+	void bookMeetingRoom(MEETINGROOM meetingRoom){
+		bookMeetingRoom(meetingRoom, 1);
+	}
+
+	void bookMeetingRoom(MEETINGROOM meetingRoom,int hours){
+		if(meetingRoom==MEETINGROOM.Nalanda){
+			if(nalandaFlag==false){
```
AavrutiDJun20: you can write like this if(meetingRoom==MEETINGROOM.Nalanda && !nalandaFlag)

src/archanaNJun20/BookMeetingRoom.java
```diff
@@ -0,0 +1,105 @@
...
+		}
+		else if(meetingRoom==MEETINGROOM.PratapGadh) {
+			if(pratapGadhFlag==false){
+				System.out.println(meetingRoom+" meeting room is booked for "+hours+"hr");
+				pratapGadhHours=hours;
+				pratapGadhFlag=true;
+			}
+			else
+				System.out.println("PratapGadh meeting room  is already booked.");
+		}
```
AavrutiDJun20, line 35: no need to write multiple else condition as you are already using if...else if ladder.
you can write one else condition at end.

src/archanaNJun20/BookMeetingRoom.java
AavrutiDJun20: here you can add condition like if all the meeting rooms are booked and give warning to user if rooms are not available

src/archanaNJun20/BookMeetingRoom.java
ArchanaNJun20: Code is updated with all review comments,

AavrutiDJun20 (approved):
Code looks good to me

PR #859, opened 2020-07-14
Anshu-Assignment-27-30min-Done

AnshuKJun20:
image
image


PranitaPJun20 (approved):
Code looks good.The output is exactly as expected.Just one suggestion you can call displayAvailableMeetingRoom() method first before calling any method and also after each room selection to check how many rooms are left now.

PR #861, opened 2020-07-14
Kartikey_Assignment_27_35min

KartikeyDJune20:
image

### DIFF
--- a/src/kartikeyDJune20/MeetingRoom.java
+++ b/src/kartikeyDJune20/MeetingRoom.java
@@ -81,7 +81,9 @@ public class MeetingRoom {
 		meetingRoom.bookmeetigRoom("Takshila", 3);
 		meetingRoom.bookmeetigRoom("agrafort");
 		meetingRoom.displayInfo();
+		System.out.println();
 		meetingRoom.availableMeetingRoom();
+		System.out.println("**********************************************");
 	}
 
 }
--- a/src/kartikeyDJune20/MeetingRoom.java
+++ b/src/kartikeyDJune20/MeetingRoom.java
@@ -1,0 +1,87 @@
+package kartikeyDJune20;
+import java.util.Scanner;
+public class MeetingRoom {
+//	Scanner scanner = new Scanner(System.in);
+	 boolean nalandaBooked = false;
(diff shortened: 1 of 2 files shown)


src/kartikeyDJune20/MeetingRoom.java
```diff
@@ -0,0 +1,87 @@
...
+			System.out.println("Takshila meeting room is avaibale");
+	}
+
+	public static void main(String[] args) {
+		MeetingRoom meetingRoom = new MeetingRoom();
+		meetingRoom.bookMeetingRoom();
+		meetingRoom.bookmeetigRoom("Takshila", 3);
+		meetingRoom.bookmeetigRoom("agrafort");
+		meetingRoom.displayInfo();
+		meetingRoom.availableMeetingRoom();
```
harshvegada, line 85: hi @KartikeyDJune20 
here displayInfo() method showing wrong information, while i m trying to book available meeting room,
after calling displayInfor() method call bookmeetingRoon("PratapGadh");
here logic is failing  little hint:

what you should do is define String array named as meetingRoomName and second array which contains hours.

int hours[] = new int[4];

KartikeyDJune20: Sure i will incorporate the changes today

src/kartikeyDJune20/MeetingRoomUsingEnum.java
```diff
@@ -0,0 +1,101 @@
...
+				System.out.println(roomName+" room you want is already booked for "+agrafortBookingTime+" hrs");
+			break;
+		case PRATAPGARDH:
+			if (pratapGardhBooked == false) {
+				pratapGardhBooked = true;
+				pratapGrardhBookingTime = time;
+			} else
+				System.out.println(roomName+" room you want is already booked for "+pratapGrardhBookingTime+" hrs");
+			break;
+		default:
```
KartikeyDJune20, line 49: Please ignore this as i forget it to comment .. we are using enum class so there is no use of it..

harshvegada (commented):
i guess this is duplicated class of PR 866

PR #864, opened 2020-07-14
ADT_Assignment_27_Done_45Mins

AditiGJUN20:
### Assignment_27
Time: 45 Mins+10 MIns(ENUM code)
Screen Shot 2020-07-14 at 11 10 17 PM

### DIFF
--- a/src/aditiGJUN20/MeetingRoom.java
+++ b/src/aditiGJUN20/MeetingRoom.java
@@ -30,7 +30,7 @@ public class MeetingRoom {
 		if (!nalandaBookingStatus) {
 			System.out.println("Nalanda Meeting room is booked for 1 hr");
 			nalandaBookingStatus = true;
-			bookedHoursNalanda=1;
+			bookedHoursNalanda = 1;
 		} else {
 			System.out.println("Nalanda Meeting room is already Booked.Try other meeting room");
 			displayAvilableMeetingRoomForBooking();
@@ -52,7 +52,8 @@ public class MeetingRoom {
 				nalandaBookingStatus = true;
 				bookedHoursNalanda = bookingHours;
 			} else {
-				System.out.println("\nMeeting room " + meetingRoom + " is already booked for " + bookedHoursNalanda + " hrs.");
+				System.out.println(
+						"\nMeeting room " + meetingRoom + " is already booked for " + bookedHoursNalanda + " hrs.");
 				displayAvilableMeetingRoomForBooking();
 			}
 		} else if (meetingRoom == MEETINGROOM.TAKSHSHILA) {
@@ -61,7 +62,8 @@ public class MeetingRoom {
 				takshshilaBookingStatus = true;
 				bookedHoursTakshashila = bookingHours;
 			} else {
-				System.out.println("\nMeeting room " + meetingRoom + " is already booked for " + bookedHoursTakshashila + " hrs.");
+				System.out.println(
+						"\nMeeting room " + meetingRoom + " is already booked for " + bookedHoursTakshashila + " hrs.");
 				displayAvilableMeetingRoomForBooking();
 			}
 		} else if (meetingRoom == MEETINGROOM.AGRAFORT) {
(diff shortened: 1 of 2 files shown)


src/aditiGJUN20/MeetingRoom.java
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	static int bookedHoursNalanda = 0, bookedHoursTakshashila = 0, bookedHoursAgraFort = 0, bookedHoursPratapGadh = 0;
+	static boolean nalandaBookingStatus, takshshilaBookingStatus, agraFortBookingStatus, pratapGadhBookingStatus;
+
+	void bookMeetingRoom() {
+		if (!nalandaBookingStatus) {
+			System.out.println("Nalanda Meeting room is booked for 1 hr");
+			nalandaBookingStatus = true;
+			bookedHoursNalanda++;
```
AmitaRJun20: here one suggestion ,as per the code when the status is true next time it will never pass the If condition. So the bookedHoursNalanda need to be mentioned as 1 hr instead of incrementing it.

src/aditiGJUN20/MeetingRoom.java
```diff
@@ -0,0 +1,134 @@
...
+
+package aditiGJUN20;
+
+public class MeetingRoom {
+
+	enum MEETINGROOM {
+		NALANDA, TAKSHSHILA, AGRAFORT, PRATAPGADH
+	}
+
+	static int bookedHoursNalanda = 0, bookedHoursTakshashila = 0, bookedHoursAgraFort = 0, bookedHoursPratapGadh = 0;
```
AmitaRJun20: here we dont need to intialize the variable as the default value is 0....

src/aditiGJUN20/MeetingRoom.java
```diff
@@ -0,0 +1,134 @@
...
+		} else {
+			System.out.println("Nalanda Meeting room is already Booked.Try other meeting room");
+			displayAvilableMeetingRoomForBooking();
+		}
+	}
+
+	void bookMeetingRoom(MEETINGROOM meetingRoom) {
+
+		if (meetingRoom == MEETINGROOM.NALANDA)
+			System.out.println("\nUse default method to book Meeting room - Nalanda");
```
AmitaRJun20: Here this if condition is not required as you have already handled the condition in 
void bookMeetingRoom(MEETINGROOM meetingRoom, int bookingHours) method...

AmitaRJun20 (changes requested):
I have mentioned some suggestions in code...please go through it...
Rest all code is good...

src/aditiGJUN20/MeetingRoom.java
AditiGJUN20: I've written this condition for requirement point B.
`b. Write a method that helps to book a meeting room other than "Nalanda" for default timing would be 1 hr.`

If we don't put the condition and if we change calling order then with this method user will be able to Book Meeting room Nalanda for one hour.
`meetingRoom.bookMeetingRoom(MEETINGROOM.NALANDA);
meetingRoom.bookMeetingRoom(MEETINGROOM.AGRAFORT);
meetingRoom.bookMeetingRoom();
meetingRoom.displayInfo();`

So, to book NALANDA user will use either of the methods-
`meetingRoom.bookMeetingRoom();
meetingRoom.bookMeetingRoom(MEETINGROOM.NALANDA, 3);
`
Let me know your view.

AmitaRJun20: That make sense...Approving this...

AmitaRJun20 (approved):
Code loos good...Approved...

src/aditiGJUN20/MeetingRoom.java
AditiGJUN20: Code updated as Suggested.

src/aditiGJUN20/MeetingRoom.java
AditiGJUN20: Changes Done.

src/aditiGJUN20/MeetingRoom.java
```diff
@@ -0,0 +1,139 @@
...
+		} else {
+			System.out.println("Nalanda Meeting room is already Booked.Try other meeting room");
+			displayAvilableMeetingRoomForBooking();
+		}
+	}
+
+	void bookMeetingRoom(MEETINGROOM meetingRoom) {
+
+		if (meetingRoom == MEETINGROOM.NALANDA)
+			System.out.println("\nUse default method to book Meeting room - Nalanda for one hour duration");
```
KrishnaKTechnocredits, line 43: If you right this condition here, it means that before this method calling you have to call bookMeetingRoom() method. What if user directly want to use this method and pass meetingRoom = Nalanada. 
So this condition not required.

PR #866, opened 2020-07-15
Barkha_Assignment_27_60mins

BarkhaPJun20:
Program to demonstrate overloading behavior.

Create a class MeetingRoom having method bookMeetingRoom which should provide
below facility:
meeting room names: Nalanda, Takshashila, AgraFort, and PratapGadh

a. If any user call method bookMeetingRoom() at that time“Nalanda” meeting room will be booked for 1 hr.
b. Write a method that helps to book a meeting room other than "Nalanda" for default timing would be 1 hr.
c. Write a method which helps to book a meeting room by giving the name of the meeting room and given hrs.
d. displayInfo() method should print booked meeting room info [i.e. meeting room,
booked for how many hrs].
e. displayAvilableMeetingRoomForBooking() method will print available meeting room names. 
     [i.e. If Nalanda and Takshashila is booked then this method should print "Available meeting rooms: AgraFort and      PratapGadh"] 
BY USING ENUM
image

### DIFF
--- a/src/barkhaPJun20/OverloadingBehaviour_MeetingRoom_enum.java
+++ b/src/barkhaPJun20/OverloadingBehaviour_MeetingRoom_enum.java
@@ -1,0 +1,91 @@
+package barkhaPJun20;
+
+public class OverloadingBehaviour_MeetingRoom_enum {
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Perfect @BarkhaPJun20 ever condition fulfilled
Keep it up

harshvegada (approved):
all changes as per the expectation
keep it up

PR #871, opened 2020-07-15
AMITA_Assignment_27_Done_40mins

AmitaRJun20:
Assignment 27 Completed...Please Review...
MeetingRoom

### DIFF
--- a/src/amitaRJun20/MeetingRoom.java
+++ b/src/amitaRJun20/MeetingRoom.java
@@ -1,0 +1,132 @@
+/*Assignment 27 : 14th July 2020
+Program to demonstrate overloading behavior.
+Create a class MeetingRoom having method bookMeetingRoom which should provide
+below facility:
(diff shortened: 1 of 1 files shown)


src/amitaRJun20/MeetingRoom.java
```diff
@@ -0,0 +1,132 @@
...
+
+public class MeetingRoom {
+	static boolean nalandaBookingStatus, takshashilaBookingStatus, agraFortBookingStatus, pratapGadhBookingStatus;
+	static double nalandaBookingHours, takshashilaBookingHours, agraFortBookingHours, pratapGadhBookingHours;
+
+	enum MEETINGROOM {
+		NALANDA, TAKSHASHILA, AGRAFORT, PRATAPGADH;
+	}
+
+	void bookMeetingRoom() {
```
MaheshKJun20, line 25: Hi Amita, 
You can directly call bookMeetingRoom(MEETINGROOM meetingRoomName, double hours) method there as we have already written this condition that method

PR #875, opened 2020-07-15
Karan assginment26 done

KARAN1595:
Completed assignment 26

### DIFF
--- a/src/kARAN1595/NumberSumfromString.java
+++ b/src/kARAN1595/NumberSumfromString.java
@@ -1,0 +1,26 @@
+package kARAN1595;
+
+public class NumberSumfromString {
+	
+	void displaySumOfNumbersFromString(String str) {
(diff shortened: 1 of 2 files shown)


src/kARAN1595/NumberSumfromString.java
```diff
@@ -0,0 +1,26 @@
...
+		int sum = 0;
+		for (int index=0;index<str.length();index++) {
+			if(Character.isDigit(str.charAt(index))) {
+				temp +=str.charAt(index);
+			}else {
+				sum+= Integer.parseInt(temp);
+				temp = "0";
+			}
+		}
+		if(!temp.equals("0"))
```
harshvegada, line 16: here length should be greater than 0, not value

harshvegada (approved):
correct logic

PR #885, opened 2020-07-16
Viresh_Assignment_28_15Mins

VireshJJun20:

image

### DIFF
--- a/src/vireshJJun20/Assignment28/sumOfNumbersAscii.java
+++ b/src/vireshJJun20/Assignment28/sumOfNumbersAscii.java
@@ -1,0 +1,33 @@
+/*
+WAP to sum all the numbers in given string, without using Character class method. 
+(Note : Use Ascii value concept) .
+
+Input: J7u9y8h1h8j4b7j3j6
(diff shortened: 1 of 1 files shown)


src/vireshJJun20/Assignment28/sumOfNumbersAscii.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class sumOfNumbersAscii {
+
+	void sumUsingAscii(String input) {
+		int sum=0;
+		boolean flag=false;
+		for (int index= 0; index < input.length(); index++) {
+			if (input.charAt(index)>=48 && input.charAt(index)<='9') { // we can compare the current index either with ASCII value or char. 
+				sum += Integer.parseInt(String.valueOf(input.charAt(index)));
+				flag= true;
```
KrishnaKTechnocredits, line 18: Why flag is needed here. If you remove it and directly print sum after for loop, you get the answer. 
sum += (str.charAt(index) - '0'); You can use this as well. 
Or  sum = sum + Character.getNumericValue(str.charAt(index);

PR #908, opened 2020-07-17
Vaishnavi_Assignment_29_25min_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20/ConverCaseUsingAscii.java
+++ b/src/vaishnaviVJun20/ConverCaseUsingAscii.java
@@ -1,5 +1,12 @@
 package vaishnaviVJun20;
 
+/*WAP to convert case in all elements of array. 
+Input : 
+String[] array = {"AaVruti", "NiKhil", "AdiTi", "MAhesh", "ArchaNA", "jaGAdeesh"};
+
+Output:
+array = {"aAvRUTI", "nIkHIL","aDItI", "maHESH","aRCHAna","JAgaDEESH"}; */
+
 public class ConverCaseUsingAscii {
 
 	String[] covertCaseInAllElements(String[] nameArray) {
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Good job @VaishnaviVJun20 
One suggestion always go with the approach which is done by exiting methods, like you use methods on Character class, if interviewer ask you different approach which not using any exiting method then go for ASCII logic. once you write this in first time then probably next next questions difficulty level will increases

but yes, this is also good one
keep it up!

src/vaishnaviVJun20/ConverCaseUsingAscii.java
```diff
@@ -0,0 +1,37 @@
...
+		}
+		return nameArray;
+	}
+
+	public static void main(String[] args) {
+		String[] array = { "VaishNAvi", "NiKhil", "AdiTi", "MAhesh", "ArchaNA", "jaGAdeesh" };
+		String[] newArray = new ConverCaseUsingAscii().covertCaseInAllElements(array);
+		System.out.print("Output array: {");
+		for (int index = 0; index < newArray.length; index++) {
+			System.out.print(newArray[index] + ", ");
```
KrishnaKTechnocredits, line 32: Viashnavi we don't know the method  Arrays.toString(newArray) to print whole array ??

VaishnaviVJun20: --No actually I didn't know about this, I will definitely keep in mind to use this method from next time. 
--Thank you @KrishnaKTechnocredits :)

VaishnaviVJun20:
> Good job @VaishnaviVJun20
> One suggestion always go with the approach which is done by exiting methods, like you use methods on Character class, if interviewer ask you different approach which not using any exiting method then go for ASCII logic. once you write this in first time then probably next next questions difficulty level will increases
> 
> but yes, this is also good one
> keep it up!


--Thank you @harshvegada I will keep this in mind :)

PR #911, opened 2020-07-17
Viresh_Assignment_29_Done_40Mins

VireshJJun20:

image

### DIFF
--- a/src/vireshJJun20/ConvertCase.java
+++ b/src/vireshJJun20/ConvertCase.java
@@ -21,16 +21,16 @@ public class ConvertCase {
 	String[] StringConvertCase(String[] inputArray) {
 		String[] outputArray = inputArray;
 		for (int arrIndex = 0; arrIndex < inputArray.length; arrIndex++) {
-			String str = inputArray[arrIndex];
+			//String str = inputArray[arrIndex]; //removed str as per reviewer suggestion
 			String tempString = "";
-			for (int strIndex = 0; strIndex < str.length(); strIndex++) {
+			for (int strIndex = 0; strIndex < inputArray[arrIndex].length(); strIndex++) {
 
-				if (str.charAt(strIndex) >= 65 && str.charAt(strIndex) <= 90) {
-					tempString += (char) ((int) str.charAt(strIndex) + 32);
-				} else if (str.charAt(strIndex) >= 'a' && str.charAt(strIndex) <= 'z') {
-					tempString += (char) ((int) str.charAt(strIndex) - 32);
+				if (inputArray[arrIndex].charAt(strIndex) >= 65 && inputArray[arrIndex].charAt(strIndex) <= 90) {
+					tempString += (char) ((int) inputArray[arrIndex].charAt(strIndex) + 32);
+				} else if (inputArray[arrIndex].charAt(strIndex) >= 'a' && inputArray[arrIndex].charAt(strIndex) <= 'z') {
+					tempString += (char) ((int) inputArray[arrIndex].charAt(strIndex) - 32);
 				} else {
-					tempString2 += str.charAt(strIndex) + " ";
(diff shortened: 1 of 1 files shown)


src/vireshJJun20/ConvertCase.java
```diff
@@ -0,0 +1,67 @@
...
+public class ConvertCase {
+
+	static String tempString2 = "";
+
+	// Method accepts Array, converts each index to appropriate case and then saves
+	// in output array.
+	String[] StringConvertCase(String[] inputArray) {
+		String[] outputArray = inputArray;
+		for (int arrIndex = 0; arrIndex < inputArray.length; arrIndex++) {
+			String str = inputArray[arrIndex];
```
DeavinaSJun20: we can remove the **str** variable here and just apply the **charAt** function on **inputArray[index].**

src/vireshJJun20/ConvertCase.java
```diff
@@ -0,0 +1,67 @@
...
+			String tempString = "";
+			for (int strIndex = 0; strIndex < str.length(); strIndex++) {
+
+				if (str.charAt(strIndex) >= 65 && str.charAt(strIndex) <= 90) {
+					tempString += (char) ((int) str.charAt(strIndex) + 32);
+				} else if (str.charAt(strIndex) >= 'a' && str.charAt(strIndex) <= 'z') {
+					tempString += (char) ((int) str.charAt(strIndex) - 32);
+				} else {
+					tempString2 += str.charAt(strIndex) + " ";
```
DeavinaSJun20: why are we using the second else. could you please explain?

src/vireshJJun20/ConvertCase.java
```diff
@@ -0,0 +1,67 @@
...
+		String[] inputArr = { "AaVruti", "NiKhil", "AdiTi", "MAhesh", "ArchaNA", "jaGAdeesh" };
+		System.out.print("Input Array : (");
+		new ConvertCase().displayArray(inputArr);
+		String[] outputArray = new ConvertCase().StringConvertCase(inputArr);
+		System.out.println("");
+		System.out.print("Output Array: (");
+		new ConvertCase().displayArray(outputArray);
+		System.out.println("");
+		if (!(tempString2.isEmpty())) {
+			System.out.println("Array contains few Strings with special chars/integers which cant be converted to either upper case/lower case.");
```
DeavinaSJun20, line 63: Also please explain the use for this loop

DeavinaSJun20 (changes requested):
Please resolve the comments

src/vireshJJun20/ConvertCase.java
VireshJJun20: Yes we can, however I had used it for more readability.

src/vireshJJun20/ConvertCase.java
VireshJJun20: This is used to capture special character Or a digit in an array element.

src/vireshJJun20/ConvertCase.java
VireshJJun20: If you are asking about the if condition, its just a display message if the Array element contains any special character or any digit.

src/vireshJJun20/ConvertCase.java
```diff
@@ -0,0 +1,67 @@
...
+package vireshJJun20;
+
+public class ConvertCase {
+
+	static String tempString2 = "";
+
+	// Method accepts Array, converts each index to appropriate case and then saves
+	// in output array.
+	String[] StringConvertCase(String[] inputArray) {
+		String[] outputArray = inputArray;
```
KrishnaKTechnocredits, line 22: If you are creating a new array then you should create till it's length String[] outputArray = new String[inputArray.length], here you are assigning old array in new then what's the difference?

PR #915, opened 2020-07-17
Mahesh_Assignment_29_15mins_Done

MaheshKJun20:
Mahesh_Assignment_29_15mins_Done

### DIFF
--- a/src/maheshKJun20/StringExamples/ConvertCase.java
+++ b/src/maheshKJun20/StringExamples/ConvertCase.java
@@ -8,14 +8,17 @@ array = {"aAvRUTI", "nIkHIL","aDItI", "maHESH","aRCHAna","JAgaDEESH"};
 
 package maheshKJun20.StringExamples;
 
+import java.util.Arrays;
+
 public class ConvertCase {
 	
 	//Method to Iterate through Array and convert case of each string element
 	void displayArraywithConvertedCase(String[] strArray) {
-		System.out.println("Output is below:");
+		String[] tempArray = new String[strArray.length];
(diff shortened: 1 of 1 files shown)


MaheshKJun20:
image

harshvegada (approved):
Good @MaheshKJun20 you wrote two function but always go with approach which is done by exiting methods, like you use methods on Character class, if interviewer ask you different approach which not using any exiting method then go for ASCII logic.

both logic is correct
keep it up!

src/maheshKJun20/StringExamples/ConvertCase.java
```diff
@@ -0,0 +1,62 @@
...
+		return newStr;
+	}
+	
+	//Optional Method to convert case of each character in String and return the converted string value. It is using Ascii value of character.
+	String convertCaseInStringUsingAscii(String str) {
+		String newStr = "";
+		for (int index=0;index<str.length();index++) {
+			int asciiValue = str.charAt(index);
+			if (asciiValue >= 65 && asciiValue <= 90)
+				newStr += Character.toLowerCase(str.charAt(index));
```
KrishnaKTechnocredits, line 46: If interviewer will ask you not to use  toLowerCase/toLowerCase method then, what you will do ?

MaheshKJun20: Hi Krishna, 
We can use Ascii value to convert case
example - asciivalue +/- 32 
will udpate the code with this logic in second function to make it purely ascii based logic.
Thanks

PR #920, opened 2020-07-17
Jagadeesh_Assignment_29_35min

JagadeeshMJun20:
Please review and comment my code.
Output:
image

### DIFF
--- a/src/jagadeeshMJun20/ConvertCaseInStringArray.java
+++ b/src/jagadeeshMJun20/ConvertCaseInStringArray.java
@@ -1,0 +1,34 @@
+/*Assignement 29 : 17th July 2020
+
+WAP to convert case in all elements of array. 
+Input : 
+String[] array = {"AaVruti", "NiKhil", "AdiTi", "MAhesh", "ArchaNA", "jaGAdeesh"};
+
+Output:
+array = {"aAvRUTI", "nIkHIL","aDItI", "maHESH","aRCHAna","JAgaDEESH"};
+ */
+package jagadeeshMJun20;
+public class ConvertCaseInStringArray {
+	void displayOutput(String[] names) {
+		String outputNames[] = { "", "", "", "", "", "" };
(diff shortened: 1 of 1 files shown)


src/jagadeeshMJun20/ConvertCaseInStringArray.java
```diff
@@ -0,0 +1,34 @@
...
+
+Output:
+array = {"aAvRUTI", "nIkHIL","aDItI", "maHESH","aRCHAna","JAgaDEESH"};
+ */
+package jagadeeshMJun20;
+public class ConvertCaseInStringArray {
+	void displayOutput(String[] names) {
+		String outputNames[] = { "", "", "", "", "", "" };
+		for (int outerIndex = 0; outerIndex < names.length; outerIndex++) {
+			for (int index = 0; index < names[outerIndex].length(); index++) {
```
MadhuraMJun20, line 15: I think we have to write this code using ASCII values. Otherwise code is good.

MadhuraMJun20 (commented):
Try to write this code using ASCII values concept.

src/jagadeeshMJun20/ConvertCaseInStringArray.java
```diff
@@ -0,0 +1,34 @@
...
+Input : 
+String[] array = {"AaVruti", "NiKhil", "AdiTi", "MAhesh", "ArchaNA", "jaGAdeesh"};
+
+Output:
+array = {"aAvRUTI", "nIkHIL","aDItI", "maHESH","aRCHAna","JAgaDEESH"};
+ */
+package jagadeeshMJun20;
+public class ConvertCaseInStringArray {
+	void displayOutput(String[] names) {
+		String outputNames[] = { "", "", "", "", "", "" };
```
KrishnaKTechnocredits, line 13: Here if you want to create new array then please write like this String outputNames[]  = new String[names.length];
String outputNames[] = { "", "", "", "", "", "" };
here problem is if I have added one more name in names then you have to assign index manually.

PR #923, opened 2020-07-17
Barkha assignment 29_50min

BarkhaPJun20:
WAP to convert case in all elements of array. 
Input : 
String[] array = {"AaVruti", "NiKhil", "AdiTi", "MAhesh", "ArchaNA", "jaGAdeesh"};

Output:
array = {"aAvRUTI", "nIkHIL","aDItI", "maHESH","aRCHAna","JAgaDEESH"};
image


harshvegada (approved):
Try to do using existing function only, because once you write this logic then next question may be difficult so try to use exiting functions for handle this kind scenarios for future. this is also good, keep it up @BarkhaPJun20 

one add on, if you can write same logic using while loop, do this at ur end

PR #928, opened 2020-07-17
Pranita_Assignment_29_Done_30mins

PranitaPJun20:
image


harshvegada (approved):
Hi @PranitaPJun20 , this good logic, but try to write exiting function on Character class methods, because once you write this logic, then next question will be difficult, If interviewer ask for second approach then go without using any exiting function then go for this which you written.

this is always good approach.
keep it up !

PR #930, opened 2020-07-17
Swapna_Assignment_29_50_min

SwapnaWJun20:
image

### DIFF
--- a/src/swapnaWJun20/ConvertCaseASCII.java
+++ b/src/swapnaWJun20/ConvertCaseASCII.java
@@ -1,0 +1,34 @@
+/*WAP to convert case in all elements of array. 
+Input : 
+String[] array = {"AaVruti", "NiKhil", "AdiTi", "MAhesh", "ArchaNA", "jaGAdeesh"};
+
+Output:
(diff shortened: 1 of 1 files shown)


src/swapnaWJun20/ConvertCaseASCII.java
```diff
@@ -0,0 +1,34 @@
...
+				if (input[outerIndex].charAt(index) >= 'A' && input[outerIndex].charAt(index) <= 'Z')
+					convertString += (char) (input[outerIndex].charAt(index) + 32);
+				else
+					convertString += (char) (input[outerIndex].charAt(index) - 32);
+			}
+			input[outerIndex] = convertString;
+			System.out.print("\"" + input[outerIndex] + "\",");
+		}
+		System.out.println("}");
+	}
```
AnupSJun20, line 26: The logic looks good. Instead of hard coding the output, you can use Arrays.toSTring. First declare an output empty array String[] array = new String[input.length] and you can add the converted strings to the index of the array. When you are having an array, to print it you can write 
Syste.out.println(Arrays.toString(array);
Hope this helps.

AnupSJun20 (approved):
Code logic is perfect but output is hardcoded. So gave a suggestion. Approving the PR.

PR #937, opened 2020-07-18
Vaishnavi_Assignment_30_15min_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20/ReplaceString.java
+++ b/src/vaishnaviVJun20/ReplaceString.java
@@ -1,0 +1,38 @@
+package vaishnaviVJun20;
+
+/* Replace second last globant with Technocredits. 
+input-> Hi Globant Hello Globant Hi Globant GM Globant Hi Globant
+
+output->  Hi Globant Hello Globant Hi Globant GM Technocredits Hi Globant*/
+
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
as per the expectation,
keep it up!

just one counter question, on same question, If interviewer ask you for 3rd globant replace with "vaishnavi"  word 
how you will handle (in your code, just think on this)
no need to answer just check at your end

src/vaishnaviVJun20/ReplaceString.java
```diff
@@ -0,0 +1,38 @@
...
+		return array;
+	}
+
+	public static void main(String[] args) {
+		String str = "Hi Globant Hello Globant Hi Globant GM Globant Hi Globant";
+		String replaceWord = "Technocredits" , sourceWord = "Globant";
+		String output = "";
+		String[] array = new ReplaceString().replaceLastSecondOccurenceOfGivenString(str, replaceWord , sourceWord);
+		for(int index=0; index<array.length;index++) {
+			output += array[index] + " ";
```
KrishnaKTechnocredits, line 32: Always try to write logic in separate method.

PR #940, opened 2020-07-18
Viresh_Assignment_30_Done_20Mins

VireshJJun20:
Please review and comment.

image

### DIFF
--- a/src/vireshJJun20/Assignment30/StringOperation.java
+++ b/src/vireshJJun20/Assignment30/StringOperation.java
@@ -1,0 +1,45 @@
+/*
+Assignement 30 : 18th July 2020
+ 
+Replace second last globant with Technocredits. 
+input-> Hi Globant Hello Globant Hi Globant GM Globant Hi Globant
+
+output->  Hi Globant Hello Globant Hi Globant GM Technocredits Hi Globant
+*/
+package vireshJJun20.Assignment30;
+
+public class StringOperation {
+
+	void replaceString(String ipStr, String word, int lastNth) {
+		int cnt = 0;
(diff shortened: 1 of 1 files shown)


src/vireshJJun20/Assignment30/StringOperation.java
```diff
@@ -0,0 +1,45 @@
...
+package vireshJJun20.Assignment30;
+
+public class StringOperation {
+
+	void replaceString(String ipStr, String word, int lastNth) {
+		int cnt = 0;
+		String[] ipArr = ipStr.split(" ");
+		if (ipStr.length() >= (word.length() * 2) + 1 && cnt == 0) {
+			ipStr = "";
+			for (int index = ipArr.length - 1; index >= 0; index--) {
```
DeavinaSJun20, line 18: Please explain the logic of if statement. It seems a bit confusing.

VireshJJun20: Here we are analyzing the output at the beginning by calculating the length of input string.
Eg: if the inpust sting= "Globant"; and if we want 2nd last copy of word Globant, then its not possible. so straight away, we can say that the given string does not contain 2 copies of Globant in given input string.

VireshJJun20 (commented):
Please check my comments.

src/vireshJJun20/Assignment30/StringOperation.java
KrishnaKTechnocredits: Logic is good but try to make it simpler, because in interview is't difficult to convince interviewer.

PR #943, opened 2020-07-18
Anshu-Assignment-30-1hr-Done

AnshuKJun20:
image


PranitaPJun20 (approved):
Getting expected output by both the approaches.Good.
You can do this in more simpler way just by iterating the array in reverse order,and replace the word when count=2.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #947, opened 2020-07-18
Pranita_Assignment_30_Done_30mins

PranitaPJun20:
image

### DIFF
--- a/src/pranitaPJun20/StringReplaceWord.java
+++ b/src/pranitaPJun20/StringReplaceWord.java
@@ -1,0 +1,35 @@
+/*
+ * Assignement 30 : 18th July 2020
+ 
+Replace second last globant with Technocredits. 
+input-> Hi Globant Hello Globant Hi Globant GM Globant Hi Globant
+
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Logic looks good @PranitaPJun20 
Keep it up, no comments
up to mark

one suggestion, you should create method which takes 3 param(1st String, 2nd word to replace)
no need to change, just think on this

src/pranitaPJun20/StringReplaceWord.java
```diff
@@ -0,0 +1,35 @@
...
+				}
+			}
+		}
+		System.out.println("\nOutput Array = " + Arrays.toString(input));
+	}
+
+	public static void main(String[] args) {
+		String input = "Hi Globant Hello Globant Hi Globant GM Globant Hi Globant";
+		String[] array = input.split(" ");
+		System.out.println("Input Array = " + Arrays.toString(array));
```
KrishnaKTechnocredits, line 31: You should always print given string as input not string array. Same for the output as well. Expected String only.

PR #948, opened 2020-07-18
Palak_Assignment30_1hr

PalakSJun20:
Hi 
I have tried using all 3 ways 
1. Iterating from last Index and print result in array format 
2. Iterating from first index and returning array  print result in string format 
3. Using String method 
 
image


AbhishekSJune2020:
code looks good to me.

KrishnaKTechnocredits:
Code looks good to me.

PR #951, opened 2020-07-18
AMITA_Assignment_30_Done_25mins

AmitaRJun20:
Assignment 30 Done..
Please Review...
ReplaceSecondLastString

### DIFF
--- a/src/amitaRJun20/ReplaceSecondLastStringInarray.java
+++ b/src/amitaRJun20/ReplaceSecondLastStringInarray.java
@@ -8,12 +8,16 @@ package amitaRJun20;
 
 public class ReplaceSecondLastStringInarray {
 	
-	void replacedSecondLastString(String input, String replaceString) {
+	String replacedSecondLastString(String input, String replaceString) {
 		int newLength = input.replaceAll("Globant", "").length();
 		int totalCount = (input.length() - newLength) / replaceString.length();
 		String outputString = "";
 		int count = 0;
 		String[] strArray = input.split(" ");
+		if(totalCount<2) {
+			System.out.println("The String has no second last " +replaceString);
+			return input;
+		}
 		for (int index = 0; index < strArray.length; index++) {
 			if (strArray[index].equals(replaceString)) {
 				count++;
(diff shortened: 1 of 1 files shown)


src/amitaRJun20/ReplaceSecondLastStringInarray.java
```diff
@@ -0,0 +1,46 @@
...
+		int newLength = input.replaceAll(replaceWord, "").length();
+		int totalCount = (input.length() - newLength) / replaceWord.length();
+		String outputString = "";
+		int count = 0;
+		String[] strArray = input.split(" ");
+		if(totalCount<2) {
+			System.out.println("The String has no second last " +replaceWord+" word");
+			return input;
+		}
+		for (int index = 0; index < strArray.length; index++) {
```
MaheshKJun20, line 21: In for you can iterate from index = strArray.length -1; index>=0; index++ This will only iterate 2 times(last and second last). This will save extra iteration in case multiple occurence.

src/amitaRJun20/ReplaceSecondLastStringInarray.java
```diff
@@ -0,0 +1,46 @@
...
+			if (strArray[index].equals(replaceWord)) {
+				count++;
+			}
+			if (count == (totalCount - 1)) {
+				strArray[index] =replacedWithWord;
+				count++;
```
MaheshKJun20, line 27: 1. instead of count++ you can break loop here as we have replaced the word we need to in string - It will avoid extra iteration
2. No need to get outputString += strArray[index]+" " --> can return String.join(" ", strArray) as the final string, after for loop is completed/breaked.

MaheshKJun20 (approved):
Hi Amita, 
Code is giving the expected output. However, I have suggested few changes in the code. 
Please have a look into them before submitting for further review. Let me know in case you need more details
Thanks

src/amitaRJun20/ReplaceSecondLastStringInarray.java
AmitaRJun20: Yes you are correct...Actually i did this to practice the way Maulik has suggested to get the totalCount in the first approach...
Thanks for approving...

src/amitaRJun20/ReplaceSecondLastStringInarray.java
AmitaRJun20: I have tried using break in the loop but in the output string is not adding further after getting the string " Technocredits" 
and giving output as "The Output String is :  Hi Globant Hello Globant Hi Globant GM " which is incorrect...for that i have used count++  to make that adjustment..

AmitaRJun20: Thank you for suggesting the string join method :) 👍

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #954, opened 2020-07-18
Anup_Assignment_30_done_20mins_PR

AnupSJun20:
Please review.
Program1

### DIFF
--- a/src/anupSJun20/StringReplace.java
+++ b/src/anupSJun20/StringReplace.java
@@ -11,24 +11,18 @@ public class StringReplace {
 		String[] newstr = str.split(" ");
 		int count = 0;
 		String finalString = "";
-		for(int index = 0;index<newstr.length;index++) {
+		for(int index=newstr.length-1;index>= 0;index--) {
 			if(newstr[index].equals(replace)) {
 				count++;
+			}
+			if(count == 2) {
+				newstr[index] = "Technocredits";
+				break;
 			}
 		}
 		if(count <2 ) {
 			System.out.println("Can't replace the second last element as there is no sufficient string");
 			return str;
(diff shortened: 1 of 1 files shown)


src/anupSJun20/StringReplace.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Arrays;
+
+public class StringReplace {
+	String findTheNewString(String str, String replace) {
+		String[] newstr = str.split(" ");
+		int count = 0;
+		String finalString = "";
+		for(int index = 0;index<newstr.length;index++) {
+			if(newstr[index].equals(replace)) {
+				count++;
```
harshvegada, line 16: Hi anup,
Line number 28 th logic can be placed here
if(newcount == 2) {
				newstr[index] = "Technocredits";
				break;
			}

So this will save your one for loop Time.

src/anupSJun20/StringReplace.java
```diff
@@ -0,0 +1,51 @@
...
+			}
+		}
+		if(count <2 ) {
```
harshvegada, line 23: This is Good

harshvegada (commented):
Do changes as per comment

src/anupSJun20/StringReplace.java
AnupSJun20: Thank you Harsh for the feedback. Updated the code.

src/anupSJun20/StringReplace.java
AnupSJun20: Thanks.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #961, opened 2020-07-19
Pranita_Assignment_31_Done_25mins

PranitaPJun20:
image


harshvegada (approved):
This also correct code
One suggestion  where you take input that time you can parse the input in Integer.parse()
If exception occur then again take input from user once user give valid input then only update count

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #969, opened 2020-07-20
Barkha_Assignmnet_31_20mins

BarkhaPJun20:
WAP to get first 3 numbers from user. Print 3 numbers and their sum.(Use exception handling concept. )
Note : User may enter wrong inputs, make sure program should not break.

Input :
10
TechnoCredits
2tech3
30
40
Output : values are 10,30,40 and total is 80
image

### DIFF
--- a/src/barkhaPJun20/ExceptionHandling_AcceptAndAddIntegers.java
+++ b/src/barkhaPJun20/ExceptionHandling_AcceptAndAddIntegers.java
@@ -9,23 +9,29 @@ public class ExceptionHandling_AcceptAndAddIntegers {
 		int countInt=0;
 		int sum=0;
 		int num;
+		int totalNumToAdd=0;
 		Scanner scanner=new Scanner(System.in);
-		System.out.println("Let us know how many numbers you want to add: ");
-		int  totalNumToAdd=scanner.nextInt();
-		while(countInt<totalNumToAdd) {                           //countInt=0; so countInt<totalNumToAdd(0,1,2). Program will run untill countInt<totalNumToAdd=fail
-			try {
-			System.out.println("Plz enter numbers you want to add: ");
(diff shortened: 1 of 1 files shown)


src/barkhaPJun20/ExceptionHandling_AcceptAndAddIntegers.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class ExceptionHandling_AcceptAndAddIntegers {
+	
+	void acceptAndAddInt() {
+		int countInt=0;
+		int sum=0;
+		int num;
+		Scanner scanner=new Scanner(System.in);
+		System.out.println("Let us know how many numbers you want to add: ");
+		int  totalNumToAdd=scanner.nextInt();
```
harshvegada: What if user give u String value, this will thrown exception

harshvegada (approved):
Do changes based on comment's.


And try same question using do while loop at ur end
No need to generate PR this.

src/barkhaPJun20/ExceptionHandling_AcceptAndAddIntegers.java
```diff
@@ -0,0 +1,37 @@
...
+	void acceptAndAddInt() {
+		int countInt=0;
+		int sum=0;
+		int num;
+		int totalNumToAdd=0;
+		Scanner scanner=new Scanner(System.in);
+		System.out.println("Let us know how many numbers you want to add: ");  //Need to handle this as well..if user enter String
+		try {
+			totalNumToAdd=scanner.nextInt();
+			while(countInt<totalNumToAdd) {                           //countInt=0; so countInt<totalNumToAdd(0,1,2). Program will run untill countInt<totalNumToAdd=fail
```
KrishnaKTechnocredits, line 17: You can use while loop instead of using variables.

PR #970, opened 2020-07-20
Vaishnavi_Assignment_31_25min_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20_ExceptionHandling/SumOfFirstThreeNumbers.java
+++ b/src/vaishnaviVJun20_ExceptionHandling/SumOfFirstThreeNumbers.java
@@ -1,0 +1,44 @@
+package vaishnaviVJun20_ExceptionHandling;
+
+import java.util.Arrays;
+import java.util.InputMismatchException;
+
+/* WAP to get first 3 numbers from user. Print 3 numbers and their sum.(Use exception handling concept. )
+Note : User may enter wrong inputs, make sure program should not break.
+
+Input :
+10
+TechnoCredits
+2tech3
+30
+40
+Output : values are 10,30,40 and total is 80 */
+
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Well handled scenarios when user not expected input
Good u used array but this will done without using array

src/vaishnaviVJun20_ExceptionHandling/SumOfFirstThreeNumbers.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Scanner;
+
+public class SumOfFirstThreeNumbers {
+	Scanner scanner = new Scanner(System.in);
+	int count = 0, sum = 0;
+	int[] array = new int[3];
+
+	void getSumOfFirstThreeNumbers() {
+		System.out.println("Please enter numbers to add: ");
+		do {
```
harshvegada, line 26: Good u used do while loop
Different than other's

VaishnaviVJun20:
> Well handled scenarios when user not expected input
> Good u used array but this will done without using array

-- Thank you @harshvegada.
--Yes, I tried without array first and it was handling the program correctly, However I wasn't able to achieve suggested output as we also  need to print which numbers were entered as input by user without using array.
-- Please suggest if there is any way to do that, I would like to practice it as well :)

src/vaishnaviVJun20_ExceptionHandling/SumOfFirstThreeNumbers.java
VaishnaviVJun20: Thank you so much @harshvegada :)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #976, opened 2020-07-20
Archana_Assignment_31_30min_Done

ArchanaNJun20:
image

### DIFF
--- a/src/archanaNJun20/PrintNumberException.java
+++ b/src/archanaNJun20/PrintNumberException.java
@@ -1,0 +1,27 @@
+package archanaNJun20;
+import java.util.InputMismatchException;
+import java.util.Scanner;
+public class PrintNumberException {
+	void sumOfDigits() {
+		int[] array = new int[3];
+		int counter = 0, sum = 0;
+		Scanner scanner = new Scanner(System.in);
+		System.out.println("Enter numbers:");
+		while(counter<=2) {
+			try {
+				array[counter] = scanner.nextInt();
+				sum += array[counter];
+				counter++;
+			}catch(InputMismatchException ie) {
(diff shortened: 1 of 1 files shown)


src/archanaNJun20/PrintNumberException.java
```diff
@@ -0,0 +1,27 @@
...
+			try {
+				array[counter] = scanner.nextInt();
+				sum += array[counter];
+				counter++;
+			}catch(InputMismatchException ie) {
+				System.out.println("Please enter integer numeber");
+				scanner.next();
+			}
+		}
+		System.out.println("Sum of the integer numbers : "+sum);
```
AavrutiDJun20: the expected output is "Output : values are 10,30,40 and total is 80" is this
we have to show the valid numbers as well for which we calculated sum.

ArchanaNJun20: code is updated as per suggestion 
image

src/archanaNJun20/PrintNumberException.java
```diff
@@ -0,0 +1,30 @@
...
+				sum += array[counter];
+				counter++;
+			}catch(InputMismatchException ie){
+				System.out.println("Please enter integer number");
+				scanner.next();
+			}
+		}
+		for(int index=0;index<array.length;index++){
+			System.out.print(array[index] + ", ");
+		}
```
AavrutiDJun20: Instead of creating one extra loop you can create one String and keep on concatenating the value like this
concatNumber += array[counter] + ", ";
and display it

AavrutiDJun20 (approved):
Code looks good to me

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #979, opened 2020-07-20
Raj_Assignment_31_Done_35mins

RajAJun20:
Assignment_31


AnshuKJun20 (approved):
Output is as expected you can also do by without using array.
int data = scanner.nextInt();
if(data>0){
sum += data
count++;
}
I think this is simple.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #982, opened 2020-07-20
Swapna_Assignment_31_Done_40_min

SwapnaWJun20:
image

### DIFF
--- a/src/swapnaWJun20/SumExceptionHandling.java
+++ b/src/swapnaWJun20/SumExceptionHandling.java
@@ -1,0 +1,37 @@
+package swapnaWJun20;
+
+import java.util.InputMismatchException;
+import java.util.Scanner;
+
+public class SumExceptionHandling {
+
+	int getSumThreeNos() {
+		int numbers = 0;
+		int sum = 0;
+		int count = 0;
+	
+		Scanner sc = new Scanner(System.in);
+		System.out.println("How many numbers you want to add : ");
+		int totalNo = sc.nextInt();	
+		while (count < totalNo) {
+			try {
+				System.out.println("Enter the number to add : ");
+				numbers = sc.nextInt();
+				sum += numbers;
+				count++;
+			} catch (InputMismatchException ie) { //
+				sc.next();
+			}
+		}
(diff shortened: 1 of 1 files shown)


src/swapnaWJun20/SumExceptionHandling.java
```diff
@@ -0,0 +1,37 @@
...
+		System.out.println("How many numbers you want to add : ");
+		int totalNo = sc.nextInt();	
+		while (count < totalNo) {
+			try {
+				System.out.println("Enter the number to add : ");
+				numbers = sc.nextInt();
+				sum += numbers;
+				count++;
+			} catch (InputMismatchException ie) { //
+				sc.next();
```
AnupSJun20: If the user add a String with space then this can't handle this. So sc.nextLine() is advisable. Also a message for the user would be good.

src/swapnaWJun20/SumExceptionHandling.java
```diff
@@ -0,0 +1,37 @@
...
+		int numbers = 0;
+		int sum = 0;
+		int count = 0;
+	
+		Scanner sc = new Scanner(System.in);
+		System.out.println("How many numbers you want to add : ");
+		int totalNo = sc.nextInt();	
+		while (count < totalNo) {
+			try {
+				System.out.println("Enter the number to add : ");
```
AnupSJun20: Here you should add the number of the integer that user is adding. So something like(count+1) in the sop statement.

AnupSJun20 (approved):
overall code looks good. But you need to make the changes. Approving for further review. Sorry for the delay in feedback.

src/swapnaWJun20/SumExceptionHandling.java
SwapnaWJun20: Suggested changes incorporated in code.

src/swapnaWJun20/SumExceptionHandling.java
SwapnaWJun20: Said changes incorporated. Thanks for comment.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #985, opened 2020-07-20
Anup_Assignment_31_Done_20mins

AnupSJun20:
Please review.
Program1

### DIFF
--- a/src/anupSJun20/SumOfThreeNumbersTryCatch.java
+++ b/src/anupSJun20/SumOfThreeNumbersTryCatch.java
@@ -1,0 +1,46 @@
+package anupSJun20;
+/*WAP to get first 3 numbers from user. Print 3 numbers and their sum.(Use exception handling concept. )
+Note : User may enter wrong inputs, make sure program should not break.
+
+Input :
+10
+TechnoCredits
+2tech3
+30
+40
+Output : values are 10,30,40 and total is 80*/
+
+import java.util.Arrays;
+import java.util.InputMismatchException;
+import java.util.Scanner;
+
+public class SumOfThreeNumbersTryCatch {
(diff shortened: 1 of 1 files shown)


src/anupSJun20/SumOfThreeNumbersTryCatch.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Arrays;
+import java.util.InputMismatchException;
+import java.util.Scanner;
+
+public class SumOfThreeNumbersTryCatch {
+	void findSumofThreeNumbers() {
+		Scanner sc = new Scanner(System.in);
+		System.out.println("Please enter the number of digits to add :");
+		int totalcount = sc.nextInt();
```
harshvegada, line 21: What if user gives String values in int
How I will handle this scenario

src/anupSJun20/SumOfThreeNumbersTryCatch.java
```diff
@@ -0,0 +1,46 @@
...
+		int count = 0;
+		int sum = 0;
+		int[] input = new int[totalcount];
+		
+		while(count<totalcount)
+			try {
+				System.out.println("Please enter the number :" + (count+1));
+				input[count] = sc.nextInt();
```
harshvegada, line 29: Without using array, u can add values

harshvegada (approved):
No Major changes required
Code looks Good to me
Keep it up 👍

src/anupSJun20/SumOfThreeNumbersTryCatch.java
KrishnaKTechnocredits: You should handle this line in try catch.

PR #993, opened 2020-07-21
Barkha_Assignment_32_15mins

BarkhaPJun20:
WAP to do sum of numbers in given statement. (Use split method and Exception handling concept. Do not use character class methods and ascii concept.) 

Input : I have 5 Years 6 month of experience.
Output : 11
image


KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1001, opened 2020-07-21
Mahesh_Assignment_32_15Mins_Done

MaheshKJun20:
image

### DIFF
--- a/src/maheshKJun20/StringExamples/SumOfDigitsInStringUsingException.java
+++ b/src/maheshKJun20/StringExamples/SumOfDigitsInStringUsingException.java
@@ -4,19 +4,24 @@ public class SumOfDigitsInStringUsingException {
 	
 	void displaySumOfDigits(String str) {
 		System.out.println("Input: " + str );
-		String[] tempArray = str.split(" ");
+		String[] tempArray = str.split("");
+		String temp ="0";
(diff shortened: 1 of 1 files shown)


src/maheshKJun20/StringExamples/SumOfDigitsInStringUsingException.java
```diff
@@ -0,0 +1,27 @@
...
+			try {
+				temp += "" + Integer.parseInt(tempArray[index]);
+			}catch(NumberFormatException ne) {
+				sum += Integer.parseInt(temp);
+				temp = "0";
+			}
+		}
+		if(!temp.equals("0"))
+			sum += Integer.parseInt(temp);
+		System.out.println("Output: Sum of Digits in statement is: " + sum);
```
KrishnaKTechnocredits, line 20: String[] statementArray = statement.split(" ");
		int sum = 0;
		for (int index = 0; index < statementArray.length; index++) {
			try {
				sum += Integer.parseInt(statementArray[index]);
			} catch (NumberFormatException numberFormatException) {

			}
		}
		System.out.println("Input  : " + statement);
		System.out.println("Output : " + sum);

mahesh you have added many extra validation. which is not required.

PR #1008, opened 2020-07-21
Archana_Assignement_32_done_20_mins

ArchanaNJun20:
output:
image

### DIFF
--- a/src/archanaNJun20/SumOfNumbers.java
+++ b/src/archanaNJun20/SumOfNumbers.java
@@ -1,0 +1,23 @@
+package archanaNJun20;
+public class SumOfNumbers{
+	void printSumOfDigits(String input){
+		int sum = 0;
+		String[] array= input.split(" ");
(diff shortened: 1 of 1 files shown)


AavrutiDJun20 (approved):
Code looks good to me

src/archanaNJun20/SumOfNumbers.java
```diff
@@ -0,0 +1,23 @@
...
+		int sum = 0;
+		String[] array= input.split(" ");
+		for (int index = 0; index < array.length;){
+			try{
+				sum += Integer.parseInt(array[index]);
+			} 
+			catch (NumberFormatException e){ 
+			}
+			finally {
+				index++;
```
KrishnaKTechnocredits, line 13: Archana finally block is not required because your exception is getting catch by catch block. So your program will not terminate. So is it okay if you do increment using for loop.

ArchanaNJun20:
Due to conflicts. closing branch.

PR #1017, opened 2020-07-22
Nikhil_Assignment_33_Done_25Min

NikhilMJun20:
-Completed and pushed Assignment # 33.
-Please review and comment.

-PFB screenshot of output of the program:-

Annotation 2020-07-22 151445

### DIFF
--- a/src/nikhilMJun20/FindItemId.java
+++ b/src/nikhilMJun20/FindItemId.java
@@ -1,0 +1,47 @@
+/*Assignment 33 : 22nd July 2020
+
+Extract ItemId from below input
+
+String input = https://www.facebook.com/technocredits/itemId=12345/
+String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
+String input = https://www.facebook.com/technocredits/itemId=12345&empid=123/
+String input = https://www.facebook.com/technocredits/empid=123/
+
+Output : 
+item id is 12345 
+        OR
+no item id present.
+*/
+
+package nikhilMJun20;
+
+public class FindItemId {
+	void displayItemId(String str, String toBeFound) {
+		String itemId = "";
+		for(int index = 0; index < str.length()-toBeFound.length(); index++) {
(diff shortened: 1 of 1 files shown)


src/nikhilMJun20/FindItemId.java
```diff
@@ -0,0 +1,47 @@
...
+no item id present.
+*/
+
+package nikhilMJun20;
+
+public class FindItemId {
+	void displayItemId(String str, String toBeFound) {
+		String itemId = "";
+		for(int index = 0; index < str.length()-toBeFound.length(); index++) {
+			if((str.substring(index, (index+toBeFound.length()))).equals(toBeFound)) {
```
VaishnaviVJun20: --No issues in the logic
-- However just a suggestion you can try this:
 int firstindex = str.indexOf("itemId");
String temp = str.substring(firstindex); 
and then proceed further it will be a little less complex :)

VaishnaviVJun20 (approved):
--Code looks good to me

src/nikhilMJun20/FindItemId.java
NikhilMJun20: @VaishnaviVJun20:
Yes, I fully agree. I will implement the suggestion.
Thanks!

src/nikhilMJun20/FindItemId.java
```diff
@@ -0,0 +1,47 @@
...
+public class FindItemId {
+	void displayItemId(String str, String toBeFound) {
+		System.out.println("\nGiven string:-\n"+str);
+		try {
+			String temp = str.substring(str.indexOf(toBeFound));
+			String itemId = "";
+			if(temp.charAt(toBeFound.length()) == '=') {
+				int index = toBeFound.length()+1;
+				while(Character.isDigit(temp.charAt(index))) {
+					itemId += temp.charAt(index);
```
KrishnaKTechnocredits, line 27: Suggesting an alternative You can also write : 

PR #1023, opened 2020-07-22
Anshu-Assignment-33-1hr-Done

AnshuKJun20:
image

### DIFF
--- a/src/anshuKJun20/ExtractItemIdFromString.java
+++ b/src/anshuKJun20/ExtractItemIdFromString.java
@@ -1,0 +1,61 @@
+package anshuKJun20;
+
+public class ExtractItemIdFromString {
+
+	void findItemId(String input) {
+	//first Apprach;
+	   /* String replaceStr = input.replace("/", "&");
+		String[] array = replaceStr.split("&");
+		int num = 0;
+		for (int index = 0; index < array.length; index++) {
+			try {
+				if (array[index].contains("itemid=12345")) {
+					String[] temparray = array[index].split("=");
+					for (int innerindex = 0; innerindex < temparray.length; innerindex++) {
+						try {
+							num += Integer.parseInt(temparray[innerindex]);
+						} catch (NumberFormatException ne) {
+						}
+					}
+				}
+			} catch (Exception e) {
+			}
+		}
+		System.out.println(" Item id is " + num);**/
+		
+		// second Approach
+		String sum = ""; int count = 0;
+		if(input.contains("itemid=")) {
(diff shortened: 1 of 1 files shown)


src/anshuKJun20/ExtractItemIdFromString.java
```diff
@@ -0,0 +1,61 @@
...
+				}
+			} catch (Exception e) {
+			}
+		}
+		System.out.println(" Item id is " + num);**/
+		
+		// second Approach
+		String sum = ""; int count = 0;
+		if(input.contains("itemid=")) {
+		int index = input.indexOf("12345");
```
PranitaPJun20: You can find substring like lastIndexOf("itemId=").Now try to find item id value without using character class method.
Either you can use replace ("itemId=", "")  or split method.

PranitaPJun20 (commented):
Please work on comments

src/anshuKJun20/ExtractItemIdFromString.java
AnshuKS20: Changes has been done.

src/anshuKJun20/ExtractItemIdFromString.java
```diff
@@ -0,0 +1,61 @@
...
+			for (int index = 0; index < array.length; index++) {
+				try {
+					number += Integer.parseInt(array[index]);
+				} catch (NumberFormatException ne) {
+
+				}
+			}
+			System.out.println("Item id in given string is: " + number);
+		} else {
+			System.out.println("Itemid is 0");
```
PranitaPJun20, line 43: Instead of this you can write "Item ID is not available"

PranitaPJun20 (approved):
Working as expected.

src/anshuKJun20/ExtractItemIdFromString.java
```diff
@@ -0,0 +1,61 @@
...
+		 **/
+
+		// second Approach
+		String number = "";
+		int count = 0;
+		if (input.contains("itemid=")) {
+			String substr = input.substring(input.lastIndexOf("itemid="), input.lastIndexOf("/"));
+			if (substr.contains("empid")) {
+				substr = substr.substring(0, substr.lastIndexOf("&"));
+			}
```
KrishnaKTechnocredits, line 32: if (input.contains("itemId=")) {
			String output = "";
					output += str.charAt(index);
				} else {
					break;   // with this code you don't need to check for empID. 
				}
			}
		} else {
		}

This code is more optimized.

PR #1024, opened 2020-07-22
Mahesh_Assignment_33_15mins_Done

MaheshKJun20:
image

### DIFF
--- a/src/maheshKJun20/StringExamples/ExtractIdFromString.java
+++ b/src/maheshKJun20/StringExamples/ExtractIdFromString.java
@@ -1,0 +1,42 @@
+/*
+Extract ItemId from below input
+String input = https://www.facebook.com/technocredits/temId=12345/
+String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Very good @MaheshKJun20 
Excellent work👌
Keep it up

src/maheshKJun20/StringExamples/ExtractIdFromString.java
```diff
@@ -0,0 +1,42 @@
...
+		if (str.contains("itemId=")) {
+			for (int index=str.indexOf("itemId=")+7;index<str.length();index++) {
+				if (String.valueOf(str.charAt(index)).equals("&") || String.valueOf(str.charAt(index)).equals("/")) {
+					break;
+				}
+				itemID += str.charAt(index);
+			}
+			System.out.println("Output: ItemId is - " + itemID);
+		}else
+			System.out.println("Output: No ItemId present in String");
```
KrishnaKTechnocredits, line 31: if (input.contains("itemId=")) {
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

PR #1027, opened 2020-07-22
Archana_Assignement_33_20mins_Done

ArchanaNJun20:
image

### DIFF
--- a/src/archanaNJun20/ExtractItemId.java
+++ b/src/archanaNJun20/ExtractItemId.java
@@ -1,12 +1,12 @@
 package archanaNJun20;
 
-public class ExtractItemId {
-	void printItemId(String input) {
+public class ExtractItemId{
+	void printItemId(String input){
 		String output = "";
 		if (input.contains("itemId")) {
-			String temp = input.substring(input.indexOf("12345"));
-			for (int index = 0; index < temp.length(); index++) {
-				if (Character.isDigit(temp.charAt(index))) {
+			String temp = input.substring(input.indexOf("itemId"));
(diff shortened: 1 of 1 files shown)


src/archanaNJun20/ExtractItemId.java
```diff
@@ -0,0 +1,29 @@
+package archanaNJun20;
+
+public class ExtractItemId {
+	void printItemId(String input) {
+		String output = "";
+		if (input.contains("itemId")) {
+			String temp = input.substring(input.indexOf("12345"));
```
AavrutiDJun20: i don't think that we can directly provide the value "12345" and extract the same value in the program.
You can substring the input till "itemId=" and than extract the Id i.e. "12345"

ArchanaNJun20: The code is updated.

AavrutiDJun20 (approved):
Code looks good to me
Good Job

src/archanaNJun20/ExtractItemId.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class ExtractItemId{
+	void printItemId(String input){
+		String output = "";
+		if (input.contains("itemId")) {
+			String temp = input.substring(input.indexOf("itemId"));
+			for (int index = 0; index < temp.length(); index++){
+				if (Character.isDigit(temp.charAt(index))){
+					output += temp.charAt(index);
+					if (!Character.isDigit(temp.charAt(index + 1)))
```
KrishnaKTechnocredits, line 11: Suggesting an alternative You can also write : 
try{

		}

PR #1029, opened 2020-07-22
Shruti assignment 33- 1.5 hours

ShrutiDJun20:
image

### DIFF
--- a/src/shrutiDJun20ExceptionHandling/ItemIDValue.java
+++ b/src/shrutiDJun20ExceptionHandling/ItemIDValue.java
@@ -10,7 +10,7 @@ public class ItemIDValue {
 		String urlafterId = "";
 		String itemId = "";
 		if (url.contains("itemId=")) {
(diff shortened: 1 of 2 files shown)


PalakSJun20 (approved):
Code Looks good to me .

src/shrutiDJun20ExceptionHandling/ItemIDValue.java
```diff
@@ -0,0 +1,45 @@
...
+												// that in future if EMP id is
+												// needed, code can be reused
+												// with minor changes
+
+		String urlafterId = "";
+		String itemId = "";
+		if (url.contains("itemId=")) {
+			urlafterId = url.substring((url.indexOf(iD) + iD.length()));
+
+			for (int index = 0; index < urlafterId.length(); index++)
```
KrishnaKTechnocredits, line 15: Good job dear, 
try{

		}

PR #1030, opened 2020-07-22
Swapna_Assignment_33_Done_50_min

SwapnaWJun20:
image

### DIFF
--- a/src/swapnaWJun20/SearchItemId.java
+++ b/src/swapnaWJun20/SearchItemId.java
@@ -1,0 +1,29 @@
+package swapnaWJun20;
+
+import javax.jws.Oneway;
+
+public class SearchItemId {
+
+	void findItemID(String input) {
+		System.out.println("\nInput : " + input);
+		if (input.contains("itemId")) {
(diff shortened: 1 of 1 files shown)


AnupSJun20 (approved):
code looks good. Getting the output. Approved.

src/swapnaWJun20/SearchItemId.java
```diff
@@ -0,0 +1,27 @@
...
+
+	void findItemID(String input) {
+		System.out.println("\nInput : " + input);
+		if (input.contains("itemId")) {
+			int itemIndex = input.lastIndexOf("itemId=");
+			int slashIndex = input.lastIndexOf("/");
+			String output = input.substring(itemIndex, slashIndex).replace("itemId=", "");
+			if (output.contains("empid")) {
+				output = output.substring(0, output.lastIndexOf("&"));
+			}
```
KrishnaKTechnocredits, line 13: if (input.contains("itemId")) {
		} else {
		}
Above is optimize solution.

try{

		}

PR #1039, opened 2020-07-22
AMITA_Assignment_33_Done_20mins

AmitaRJun20:
Assignment 33 Done...Please Review...
Capture

### DIFF
--- a/src/amitaRJun20/ExtractItemIdFromString.java
+++ b/src/amitaRJun20/ExtractItemIdFromString.java
@@ -1,0 +1,30 @@
+package amitaRJun20;
+
+public class ExtractItemIdFromString {
+	void display(String input) {
+		System.out.println("The input String is :"+input);
+		String itemId = "";
+		if (input.contains("itemId")) {
+			String temp = input.substring(input.indexOf("itemId"));
+			for (int index = 0; index < temp.length(); index++) {
+				if (Character.isDigit(temp.charAt(index))) {
+					itemId += temp.charAt(index);
+					if (!Character.isDigit(temp.charAt(index + 1))) {
+						break;
+					}
+				}
+			}
+			System.out.println("ItemId is " + itemId);
+			System.out.println();
+		} else {
+			System.out.println("ItemId is not present");
+		}
+	}
+	public static void main(String[] args) {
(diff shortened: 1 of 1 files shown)


MaheshKJun20 (approved):
Hi Amita,

Code looks good to me. Approving for further review.

src/amitaRJun20/ExtractItemIdFromString.java
```diff
@@ -0,0 +1,40 @@
...
+		System.out.println("The input String is :" + input);
+		String itemId = "";
+		if (input.contains("itemId")) {
+			String temp = input.substring(input.indexOf("itemId"));
+			for (int index = 0; index < temp.length(); index++) {
+				if (Character.isDigit(temp.charAt(index))) {
+					itemId += temp.charAt(index);
+					if (!Character.isDigit(temp.charAt(index + 1))) {
+						break;
+					}
```
KrishnaKTechnocredits, line 24: if (input.contains("itemId=")) {
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

src/amitaRJun20/ExtractItemIdFromString.java
```diff
@@ -0,0 +1,40 @@
...
+
+	void display(String input) {
+		System.out.println("The input String is :" + input);
+		String itemId = "";
+		if (input.contains("itemId")) {
+			String temp = input.substring(input.indexOf("itemId"));
+			for (int index = 0; index < temp.length(); index++) {
+				if (Character.isDigit(temp.charAt(index))) {
+					itemId += temp.charAt(index);
+					if (!Character.isDigit(temp.charAt(index + 1))) {
```
KrishnaKTechnocredits, line 22: Here instead of if block you can directly write else block for parent if.

src/amitaRJun20/ExtractItemIdFromString.java
```diff
@@ -0,0 +1,40 @@
...
+						break;
+					}
+				}
+			}
+			System.out.println("ItemId is " + itemId);
+			System.out.println();
+		} else {
+			System.out.println("ItemId is not present");
+		}
+	}
```
KrishnaKTechnocredits, line 32: Suggesting an alternative You can also write : 
try{

		}

PR #1041, opened 2020-07-23
Anup_Assignment_33_Done_30mins_PR

AnupSJun20:
Please review
Program1

### DIFF
--- a/src/anupSJun20/FindItemIDfromAString.java
+++ b/src/anupSJun20/FindItemIDfromAString.java
@@ -1,0 +1,42 @@
+package anupSJun20;
+/*Extract ItemId from below input
+
+String input = https://www.facebook.com/technocredits/temId=12345/
+String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
+String input = https://www.facebook.com/technocredits/itemId=12345&empid=123/
+String input = https://www.facebook.com/technocredits/empid=123/
+
+Output : 
+item id is 12345 
+        OR
+no item id present.*/
+
+public class FindItemIDfromAString {
+	void findItemID(String str) {
+		System.out.println("\nGiven String is : "+str);
+		String temp = "";
+		try {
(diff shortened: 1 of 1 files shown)


src/anupSJun20/FindItemIDfromAString.java
```diff
@@ -0,0 +1,42 @@
...
+        OR
+no item id present.*/
+
+public class FindItemIDfromAString {
+	void findItemID(String str) {
+		System.out.println("\nGiven String is : "+str);
+		String temp = "";
+		try {
+			String newstr = str.substring(str.indexOf("item"));
+			if (newstr.contains("item")) {
```
harshvegada, line 20: Here contains should be itemID

src/anupSJun20/FindItemIDfromAString.java
```diff
@@ -0,0 +1,42 @@
...
+				for (int index = 0; index < newstr.length(); index++) {
```
harshvegada, line 21: If condition is correct
One change is, in for loop we can starts from newstr.Indexof("item" )+4 this will return ASCII value
4 for item length, 4 for we want to skip to check "item" word

Rest logic is correct

harshvegada (commented):
Work on code comments

harshvegada (approved):

src/anupSJun20/FindItemIDfromAString.java
```diff
@@ -0,0 +1,42 @@
...
+							|| String.valueOf(newstr.charAt(index)).equals("/")) {
+						break;
+					}
+					temp += newstr.charAt(index);
+				}
+				System.out.println(temp);
+			}
+		} catch (Exception e) {
+			System.out.println("No item id present");
+		}
```
KrishnaKTechnocredits, line 32: Suggesting an alternative You can also write : 
try{

		}

PR #1050, opened 2020-07-23
Vaishnavi_Assignment_33_45min_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20_ExceptionHandling/Extract.java
+++ b/src/vaishnaviVJun20_ExceptionHandling/Extract.java
@@ -1,0 +1,44 @@
+package vaishnaviVJun20_ExceptionHandling;
+
+/* Extract ItemId from below input
+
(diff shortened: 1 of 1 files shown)


src/vaishnaviVJun20_ExceptionHandling/Extract.java
```diff
@@ -0,0 +1,44 @@
...
+
+	void getItemIdFromGivenString(String str) {
+		try {
+			int firstindex = str.indexOf("itemId");
+			String temp = str.substring(firstindex);
+			String itemID = "";
+			for (int index = 0; index < temp.length(); index++) {
+				if (Character.isDigit(temp.charAt(index))) {
+					itemID += temp.charAt(index);
+					if (!Character.isDigit(temp.charAt(index + 1))) {
```
KrishnaKTechnocredits, line 25: Suggesting an alternative You can also write : 
try{

		}

PR #1051, opened 2020-07-23
Palak assingment33 new

PalakSJun20:
image

### DIFF
--- a/src/palakSJun20/ExtractItemIdFromUrl.java
+++ b/src/palakSJun20/ExtractItemIdFromUrl.java
@@ -1,0 +1,50 @@
+/*Extract ItemId from below input
+
+	String input = https://www.facebook.com/technocredits/temId=12345/
+	String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
+	String input = https://www.facebook.com/technocredits/itemId=12345&empid=123/
+	String input = https://www.facebook.com/technocredits/empid=123/
+
+	Output : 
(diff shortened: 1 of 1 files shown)


src/palakSJun20/ExtractItemIdFromUrl.java
```diff
@@ -0,0 +1,50 @@
...
+package palakSJun20;
+
+public class ExtractItemIdFromUrl {
+	void findIdFromUrl(String input) {
+		String output = "";
+		if (input.contains("itemId")) {
+			for (int index = input.indexOf("itemId"); index < input.length(); index++) {
+				if (input.charAt(index) >= '0' && input.charAt(index) <= '9') {
+					output += input.charAt(index);
+					if (!(input.charAt(index + 1) >= '0' && input.charAt(index) <= '9')) {
```
KrishnaKTechnocredits, line 21: Here if is not required, you can write else block for parent if.

			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

src/palakSJun20/ExtractItemIdFromUrl.java
```diff
@@ -0,0 +1,50 @@
...
+						break;
+					}
+				}
+			}
```
KrishnaKTechnocredits, line 25: Suggesting an alternative You can also write : 
try{

		}

PR #1052, opened 2020-07-23
Aavruti_Assignment_33_10mins

AavrutiDJun20:
Program to Extract ItemId

image

### DIFF
--- a/src/aavrutiDJun20/ExtractItemId.java
+++ b/src/aavrutiDJun20/ExtractItemId.java
@@ -1,0 +1,31 @@
+package aavrutiDJun20;
+
+public class ExtractItemId {
+
+	//Extract ItemId
+	void findItemId(String input) {
+		System.out.println("Input  --> " + input);
+		if(input.contains("itemId")) {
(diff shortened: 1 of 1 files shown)


src/aavrutiDJun20/ExtractItemId.java
```diff
@@ -0,0 +1,31 @@
...
+	//Extract ItemId
+	void findItemId(String input) {
+		System.out.println("Input  --> " + input);
+		if(input.contains("itemId")) {
+			String itemIdValue = input.substring(input.lastIndexOf("itemId="),input.lastIndexOf("/")).replace("itemId=", "");
+			if(itemIdValue.contains("&")) {
+				itemIdValue = itemIdValue.substring(0,itemIdValue.lastIndexOf("&"));
+			}
+			System.out.println("Output --> ItemId is " + itemIdValue);
+		}
```
KrishnaKTechnocredits, line 14: if (input.contains("itemId")) {
		} else {
		}

try{

		}

PR #1053, opened 2020-07-23
Vireshj_Assignment_33_20Mins

VireshJJun20:

image

### DIFF
--- a/src/vireshJJun20/assignment33/StringOpsItemId.java
+++ b/src/vireshJJun20/assignment33/StringOpsItemId.java
@@ -1,0 +1,37 @@
+package vireshJJun20.assignment33;
+
+public class StringOpsItemId {
+
+	void getSubString(String ipStr) {
+		String tempStr1 = "", tempStr2 = "";
+		for (int index = 0; index < ipStr.length(); index++) {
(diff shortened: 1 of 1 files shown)


src/vireshJJun20/assignment33/StringOpsItemId.java
```diff
@@ -0,0 +1,37 @@
...
+				}
+			}
+		}
+		if (tempStr1.isEmpty() || !(tempStr1.equalsIgnoreCase("itemid"))) {
+			System.out.println(ipStr);
+			System.out.println("No item id present");
+		} else if (tempStr1.equalsIgnoreCase("itemid")) {
+			System.out.println(ipStr);
+			System.out.println(tempStr1.substring(0, tempStr1.length() - 2) + " "
+					+ tempStr1.substring(tempStr1.length() - 2) + " is " + tempStr2.substring(1, tempStr2.length()));
```
KrishnaKTechnocredits, line 26: Lots of complexity. 

1st way :
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

2nd way:
if (input.contains("itemId")) {
		} else {
		}

3rd and optimize way:

try{

		}

PR #1054, opened 2020-07-23
ADT_Assignment_33_Done_20Mins

AditiGJUN20:
### Assignment: 33
**Time:** 20 Mins
Screen Shot 2020-07-22 at 3 12 26 PM

### DIFF
--- a/src/aditiGJUN20/ExtractValueFromInput.java
+++ b/src/aditiGJUN20/ExtractValueFromInput.java
@@ -1,0 +1,49 @@
+/*
+ * Extract ItemId from below input
+ * String input = https://www.facebook.com/technocredits/itemId=12345/
+ * String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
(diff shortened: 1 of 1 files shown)


src/aditiGJUN20/ExtractValueFromInput.java
```diff
@@ -0,0 +1,49 @@
...
+public class ExtractValueFromInput {
+	void findValueOf(String url, String findValueOf) {
+		String itemId = "";
+		if (url.contains(findValueOf)) {
+			System.out.println("Input Url: " + url);
+			String tempStr = url.substring(url.lastIndexOf(findValueOf));
+			for (int index = 0; index < tempStr.length(); index++) {
+				if (String.valueOf(tempStr.charAt(index)).equals("&")
+						|| String.valueOf(tempStr.charAt(index)).equals("/")) {
+					break;
```
KrishnaKTechnocredits, line 27: Suggesting an alternative You can also write : 
try{

		}

PR #1056, opened 2020-07-23
Barkha_Assignment_33_45mins

BarkhaPJun20:
Extract ItemId from below input

String input = https://www.facebook.com/technocredits/temId=12345/
String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
String input = https://www.facebook.com/technocredits/itemId=12345&empid=123/
String input = https://www.facebook.com/technocredits/empid=123/

Output : 
item id is 12345 
        OR
no item id present.
image

### DIFF
--- a/src/barkhaPJun20/ExtractNumFromURL.java
+++ b/src/barkhaPJun20/ExtractNumFromURL.java
@@ -1,0 +1,37 @@
+package barkhaPJun20;
+
+public class ExtractNumFromURL {
+	String output="";
+	
+	void displayItemId(String URL) {
+		System.out.println("Input is=> "+URL);
+		String removeFB=URL.replace("https://www.facebook.com/technocredits/", "");
+		if (removeFB.contains("itemId=")) {
+			if (removeFB.contains("itemId=") && !removeFB.contains("empid=123")) {
+				for (int index=0; index<removeFB.length(); index++) {
+					char ch=removeFB.charAt(index);
+					if (Character.isDigit(ch))
+						output+=ch;
+				}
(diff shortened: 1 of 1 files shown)


src/barkhaPJun20/ExtractNumFromURL.java
```diff
@@ -0,0 +1,37 @@
+package barkhaPJun20;
+
+public class ExtractNumFromURL {
+	String output="";
+	
+	void displayItemId(String URL) {
```
harshvegada, line 6: This gives correct output @BarkhaPJun20 
But there other function as well
Like index of(), substring()

This is also recommend option which write

harshvegada (approved):
No logical changes
Well done

src/barkhaPJun20/ExtractNumFromURL.java
```diff
@@ -0,0 +1,37 @@
...
+					if (Character.isDigit(ch))
+						output+=ch;
+				}
+				System.out.println("itemID is=> "+output);
+			}else if(removeFB.contains("itemId=") && removeFB.contains("empid=123")) {  //will work as if
+				String removeEmpid=removeFB.replace("empid=123", "");
+				for (int index1=0; index1<removeEmpid.length(); index1++) {
+					char ch=removeEmpid.charAt(index1);
+					if (Character.isDigit(ch))
+						output+=ch;
```
KrishnaKTechnocredits, line 22: Your code is not optimize:
Try below solution-
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

try{

		}

PR #1057, opened 2020-07-23
Madhura_Assignment_33_45mins

MadhuraMJun20:
ExtractItemIdFromString


src/madhuraMJun20/assignment30Plus/ExtractItemIdFromString.java
```diff
@@ -0,0 +1,37 @@
...
+		if (input.contains("itemId")) {
+			for (int index = input.indexOf("itemId=") + 7; index < input.length(); index++) {
+				if (String.valueOf(input.charAt(index)).equals("&")
+						|| String.valueOf(input.charAt(index)).equals("/")) {
+					break;
+				}
+				output += input.charAt(index);
+			}
+			System.out.println("Item Id is : " + output);
+		} else {
```
KrishnaKTechnocredits, line 25: Suggesting an alternative You can also write : 
try{

		}

PR #1058, opened 2020-07-23
Pranita_Assignment_33_1_Done_45mins

PranitaPJun20:
image

### DIFF
--- a/src/pranitaPJun20/StringExtractItemID.java
+++ b/src/pranitaPJun20/StringExtractItemID.java
@@ -1,0 +1,55 @@
+/*/*Assignment 33 : 22nd July 2020
+
+	Extract ItemId from below input
+
+	String input = https://www.facebook.com/technocredits/itemId=12345/
+	String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Very Good @PranitaPJun20 
Keep it up

src/pranitaPJun20/StringExtractItemID.java
```diff
@@ -0,0 +1,55 @@
...
+		if (input.contains("itemId")) {
+			String subInput = input.substring(input.lastIndexOf("itemId="), input.lastIndexOf("/"));
+			if (subInput.contains("empid")) {
+				subInput = subInput.substring(0, subInput.lastIndexOf("&"));
+			}
+			String[] array = subInput.split("");
+			for (int index = 0; index < array.length; index++) {
+				try {
+					number += Integer.parseInt(array[index]);
+				} catch (NumberFormatException nfe) {
```
KrishnaKTechnocredits, line 32: Suggesting an alternative You can also write : 
try{

		}

Second solution.
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

PR #1059, opened 2020-07-23
Rachana_Assignment_33_Done_30min

RachanaGJun20:
image

### DIFF
--- a/src/rachanaGJun20/ExtractIteamId.java
+++ b/src/rachanaGJun20/ExtractIteamId.java
@@ -1,0 +1,51 @@
+/*Assignment 33 : 22nd July 2020
+
+Extract ItemId from below input
+
+String input = https://www.facebook.com/technocredits/itemId=12345/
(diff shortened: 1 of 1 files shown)


src/rachanaGJun20/ExtractIteamId.java
```diff
@@ -0,0 +1,51 @@
...
+			int strindex = str1.length;
+			String str2 = str1[strindex - 1];
+			if (str2.contains("&")) {
+				String[] str3 = str2.split("&");
+				for (int index = 0; index < str3.length; index++) {
+					if (str3[index].contains("itemId=")) {
+					int id = Integer.parseInt(str3[index].replace("itemId=", "").trim());
+						System.out.println("Output: Itemid is : " + id);
+					}
+				}
```
KrishnaKTechnocredits, line 31: if (input.contains("itemId=")) {
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

try{

		}

PR #1060, opened 2020-07-23
Aastha_Assignment_33_Done_30mins_2

AasthaJJun20:
-> Assignment-33 Done (created new branch since the old one had issues)
->PFB output:
ItemIDFromURL-output

### DIFF
--- a/src/aasthaJJun20/ItemIDFromURL.java
+++ b/src/aasthaJJun20/ItemIDFromURL.java
@@ -1,0 +1,53 @@
+package aasthaJJun20;
+/*Assignment 33 : 22nd July 2020
+
+Extract ItemId from below input
+
+String input = https://www.facebook.com/technocredits/itemId=12345/
+String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
+String input = https://www.facebook.com/technocredits/itemId=12345&empid=123/
+String input = https://www.facebook.com/technocredits/empid=123/
+
+Output : 
+item id is 12345 
+        OR
+no item id present.*/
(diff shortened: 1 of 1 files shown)


NikhilMJun20 (approved):
Code and logic look good to me.
I am approving the PR.

src/aasthaJJun20/ItemIDFromURL.java
```diff
@@ -0,0 +1,53 @@
...
+Output : 
+item id is 12345 
+        OR
+no item id present.*/
+public class ItemIDFromURL {
+	void findItemIDFromURL(String url) {
+		System.out.println("\nString input : " + url);
+		String temp = "";
+		try {
+			temp = url.substring(url.indexOf("itemId"));
```
KrishnaKTechnocredits, line 20: try{

		}

src/aasthaJJun20/ItemIDFromURL.java
```diff
@@ -0,0 +1,53 @@
...
+		} catch (StringIndexOutOfBoundsException stringIndexOutOfBoundsException) {
+			System.out.println("No item id present");
+			return;
+		}
+		String itemId = "";
+		for (int index = 0; index < temp.length(); index++) {
```
KrishnaKTechnocredits, line 26: if (input.contains("itemId=")) {
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

AasthaJJun20: Thanks Krishna!!

PR #1061, opened 2020-07-23
Raj_Assignment_33_Done

RajAJun20:
Assignment_33

### DIFF
--- a/src/rajAJun20/ExtractItemIdFromInput.java
+++ b/src/rajAJun20/ExtractItemIdFromInput.java
@@ -1,0 +1,37 @@
+/* Extract ItemId from below input :
+	 String input = https://www.facebook.com/technocredits/itemId=12345/
+	 String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
+	 String input = https://www.facebook.com/technocredits/itemId=12345&empid=123/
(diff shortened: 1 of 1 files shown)


AnshuKS20 (approved):
Output is as expected.

AnshuKJun20 (approved):
Output is as expected.

src/rajAJun20/ExtractItemIdFromInput.java
```diff
@@ -0,0 +1,37 @@
...
+public class ExtractItemIdFromInput {
+
+	void findItemIdFromInput(String input) {
+		String itemId = "";
+		if (input.contains("itemId")) {
+			String temp = input.substring(input.indexOf("itemId"));
+			for (int index = 0; index < temp.length(); index++) {
+				if (temp.charAt(index) >='0' && temp.charAt(index) <='9') {
+					itemId += temp.charAt(index);
+					if (!(temp.charAt(index + 1) >='0' && temp.charAt(index+1) <='9'))
```
KrishnaKTechnocredits, line 19: This if is not required, you should write else of parent if.
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

try{

		}

PR #1062, opened 2020-07-23
Jagadeesh__Assignment__33_25mins

JagadeeshMJun20:
image

### DIFF
--- a/src/jagadeeshMJun20/ExtractItemId.java
+++ b/src/jagadeeshMJun20/ExtractItemId.java
@@ -1,0 +1,38 @@
+/*Extract ItemId from below inpu9t
+String input = https://www.facebook.com/technocredits/temId=12345/
+String input = https://www.facebook.com/technocredits/empid=123&itemId=12345/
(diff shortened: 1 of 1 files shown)


src/jagadeeshMJun20/ExtractItemId.java
```diff
@@ -0,0 +1,38 @@
...
+package jagadeeshMJun20;
+ class ExtractItemID {
+	void displayItemID(String input) {
+		String itemId = "";
+		if (input.contains("itemId")) {
+			String temp1 = input.substring(input.indexOf("itemId"));
+			for (int index = 0; index < temp1.length(); index++) {
+				if (temp1.charAt(index) >='0' && temp1.charAt(index) <='9') {
+					itemId += temp1.charAt(index);
+					if (!(temp1.charAt(index + 1) >='0' && temp1.charAt(index+1) <='9'))
```
KrishnaKTechnocredits, line 19: This if is not required, you should write else of parent of this block.

			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

try{

		}

PR #1063, opened 2020-07-24
Piyush_Assignment_33_35Min

PiyushPJun20:
image

### DIFF
--- a/src/piyushPJun20/ItemId.java
+++ b/src/piyushPJun20/ItemId.java
@@ -1,0 +1,31 @@
+package piyushPJun20;
+
+public class ItemId {
+	void getItemID(String input) {
+		String itemId = "";
+		if (input.contains("itemId")) {
+			String temp = input.substring(input.indexOf("itemId"));
+			for (int index = 0; index < temp.length(); index++) {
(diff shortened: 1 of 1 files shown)


ShrutiDJun20 (approved):
code looks good to me

src/piyushPJun20/ItemId.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class ItemId {
+	void getItemID(String input) {
+		String itemId = "";
+		if (input.contains("itemId")) {
+			String temp = input.substring(input.indexOf("itemId"));
+			for (int index = 0; index < temp.length(); index++) {
+				if (temp.charAt(index) >= '0' && temp.charAt(index) <= '9') {
+					itemId += temp.charAt(index);
+					if (!(temp.charAt(index + 1) >= '0' && temp.charAt(index + 1) <= '9'))
```
KrishnaKTechnocredits, line 11: This if is not required. You should write else of above if block.
			String output = "";
					output += str.charAt(index);
				} else {
					break;
				}
			}
		} else {
		}

try{

		}

PR #1119, opened 2020-07-26
Mahesh_Coding_Exam_7_18Mins

MaheshKJun20:
image

### DIFF
--- a/src/maheshKJun20/BankUser.java
+++ b/src/maheshKJun20/BankUser.java
@@ -1,0 +1,62 @@
+package maheshKJun20;
+
+public class BankUser {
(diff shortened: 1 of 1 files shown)


src/maheshKJun20/BankUser.java
```diff
@@ -0,0 +1,62 @@
...
+	int userId = 0, userDebit =0, userCredit =0;
+	
+	// Block can be used for maintaining user count
+//	{
+//		user++;
+//	}
+	
+	BankUser(){
+		user++;
+		userId = user;
```
KrishnaKTechnocredits, line 14: This operation is not required.

PR #1136, opened 2020-07-26
Palak_Coding_Exam7_50Min

PalakSJun20:
image

### DIFF
--- a/src/palakSJun20/User.java
+++ b/src/palakSJun20/User.java
@@ -1,0 +1,81 @@
+/*Exam-7:
+Q1) class User{
+void displayInfo(){
+debit
+credit
+}
+
+void displayAllInfo(){
+totaldebit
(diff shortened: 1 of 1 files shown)


src/palakSJun20/User.java
```diff
@@ -0,0 +1,81 @@
...
+		totalDebitCount++;
+	}
+	
+	void credit() {
+		userCreditCount++;
+		totalCreditCount++;
+	}
+	
+	void displayInfo() {
+		System.out.println(toString() + " --> Debit count is " + userDebitCount);
```
KrishnaKTechnocredits, line 48: instead of toString(), cant we directly say name.
toString method should be override when you want to print User detail when you print referece.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1139, opened 2020-07-26
Aastha_Coding_Exam_7_Done_50mins

AasthaJJun20:
-> Created class UsersOfBank with all the methods and private variable balance
-> Created a custom exception to handle the case -  if the entered amount is invalid e.g. -100
-> Create client class to call the methods
-> PFB output
UsersOfBank-output

PS : took time, as I was not able to figure out why the total user are 4 but then I found that there are 2 more objects I have created in try catch with invalid amount.


KrishnaKTechnocredits (approved):
Impressive. Good job dear.

PR #1185, opened 2020-07-27
Archana_Assignement_34_Done_35mins

ArchanaNJun20:
Program 1 : 
Remove all Techno from given ArrayList.
image

Program 2 :
Display on which index Techno is available, Replace all Techno with Credits from given ArrayList, and count total number of Credits.
image

### DIFF
--- a/src/archanaNJun20/RemoveFromArrayList.java
+++ b/src/archanaNJun20/RemoveFromArrayList.java
@@ -1,0 +1,25 @@
+package archanaNJun20;
+
+import java.util.ArrayList;
+
+public class RemoveFromArrayList {
+	void removeTechno(ArrayList<String> input){
+		for(int index=0;index<input.size();index++){
+			if(input.contains("Techno"))
+				input.remove("Techno");
+		}
+		System.out.println("output String :"+input);
+	}
+	public static void main(String[] args){
+		RemoveFromArrayList removeFromArrayList=new RemoveFromArrayList();
+		ArrayList<String> input= new ArrayList<String>();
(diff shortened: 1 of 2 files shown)


src/archanaNJun20/RemoveFromArrayList.java
```diff
@@ -0,0 +1,25 @@
+package archanaNJun20;
+
+import java.util.ArrayList;
+
+public class RemoveFromArrayList {
+	void removeTechno(ArrayList<String> input){
+		for(int index=0;index<input.size();index++){
```
AavrutiDJun20: You can also directly write "for(;input.contains("Techno");)" like this

src/archanaNJun20/ReplaceTechnoWithCredits.java
```diff
@@ -0,0 +1,39 @@
...
+				System.out.print(input.indexOf("Techno") + ",");
+				input.set(input.indexOf("Techno"), "Credits");
+			}
+		}
+		System.out.println();
+		return input;
+	}
+	void countWord(ArrayList<String> input){
+		int count=0;
+		for (int index = 0; index <= input.size(); index++){
```
AavrutiDJun20, line 23: here also you can write like this "while(input.contains("Credits"))", as well as in above method "replaceTechnoWithCredits(ArrayList<String> input)"

AavrutiDJun20 (approved):
Good

KrishnaKTechnocredits (approved):
Good Job dear.

PR #1188, opened 2020-07-28
Barkha_Assignment_35_10mins

BarkhaPJun20:
String[] name1 = {"Palak", "Viresh", "Yash", "Aavruti"};
String[] name2 = {"Deavina","Palak","Viresh", "Nikhil"};
output : [Palak, Viresh, Yash, Aavruti,Deavina,Nikhil]
image


harshvegada (approved):
Very good Barkha
Keep doing well

KrishnaKTechnocredits (approved):
I am happy, you could figure out arrayList2.removeAll(arrayList1) is required and asList method returns List which could not be modified, hence we need to create new ArrayList object and call asList method from constructor.
Code looks good to me.

PR #1197, opened 2020-07-28
Anshu-Assignment-35-15min

AnshuKJun20:
image


PranitaPJun20 (approved):
Working as expected.
You can do this particular example without for loop also.By using removeAll and addAll method.

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1199, opened 2020-07-28
AMITA_Assignment_35_Done_10mins

AmitaRJun20:
Please review Assignment-35
RetriveUniqueNames


MaheshKJun20 (approved):
Hi Amita,

Code works fine for given set of input.
This solution may not work when we have duplicate value in a list. Can think on that as well.

Thanks.

AmitaRJun20:
> Hi Amita,
> 
> Code works fine for given set of input.
> This solution may not work when we have duplicate value in a list. Can think on that as well.
> 
> Thanks.

@MaheshKJun20 , thank you for the suggestion...I updated the code ...

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1203, opened 2020-07-28
Piyush_Assignment_34

PiyushPJun20:
image

image

### DIFF
--- a/src/piyushPJun20/ArrayListMethod.java
+++ b/src/piyushPJun20/ArrayListMethod.java
@@ -1,0 +1,48 @@
+package piyushPJun20;
+
+import java.util.ArrayList;
+
+public class ArrayListMethod {
+	void display(ArrayList<String> name) {
+		System.out.print("Techno word index in given arrayList is ");
+		for (int index = 0; index < name.size(); index++) {
+			if (name.get(index) == "Techno") {
+				System.out.print(index + ",");
+			}
+		}
+	}
+
+	void replaceWord(ArrayList<String> name) {
+		System.out.println();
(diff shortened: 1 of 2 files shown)


src/piyushPJun20/ArrayListMethod.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class ArrayListMethod {
+	void display(ArrayList<String> name) {
+		System.out.print("Techno word index in given arrayList is ");
+		for (int index = 0; index < name.size(); index++) {
+			if (name.get(index) == "Techno") {
+				System.out.print(index + ",");
+			}
+		}
+	}
```
ShrutiDJun20, line 13: Code looks good to me. You can also combine all the three methods in one. But this way also its correct because if user just wants to know the count he can directly call the count method.

src/piyushPJun20/RemoveArrayList.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.ArrayList;
+
+public class RemoveArrayList {
+	void remove(ArrayList<String> name) {
+		for (int index = 0; index < name.size(); index++) {
+			if (name.get(index) == ("Techno")) {
+				name.remove(index);
+			}
+		}
+		System.out.println(name);
```
ShrutiDJun20, line 12: Looks good.  As an alternative , you can just write one line in the for loop " name.remove(name.indexOf("Techno")) " , if you want to avoid the if statement.

ShrutiDJun20 (approved):

KrishnaKTechnocredits (approved):
Good job.

PR #1208, opened 2020-07-28
Swapna_Assignment_35_Done_30_min

SwapnaWJun20:
image

### DIFF
--- a/src/swapnaWJun20/RetrieveUniqueNameArray.java
+++ b/src/swapnaWJun20/RetrieveUniqueNameArray.java
@@ -1,0 +1,36 @@
+/*Assignment 35:  Retrieve all unique names from given array.
+
+String[] name1 = {"Palak", "Viresh", "Yash", "Aavruti"};
+String[] name2 = {"Deavina","Palak","Viresh", "Nikhil"};
+output : [Palak, Viresh, Yash, Aavruti,Deavina,Nikhil]*/
+
+package swapnaWJun20;
+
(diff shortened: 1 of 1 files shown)


src/swapnaWJun20/RetrieveUniqueNameArray.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class RetrieveUniqueNameArray {
+
+	ArrayList<String> getUniqueElement(ArrayList<String> alName1, ArrayList<String> alName2) {
+		ArrayList<String> tempList = new ArrayList<String>(alName1);
+		ArrayList<String> outputList = new ArrayList<String>();
+		outputList.addAll(alName2);
```
AnupSJun20, line 26: Here If we have a duplicate value in 2nd list then it will display both the duplicate list. Other wise code looks good.

AnupSJun20 (approved):
Please handle the duplicate elements in 2nd list. Otherwise code looks good.

src/swapnaWJun20/RetrieveUniqueNameArray.java
SwapnaWJun20: Thanks for the comment Anup, I have added the logic and updated the code

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1217, opened 2020-07-31
Barkha_Assignment_36_15mins

BarkhaPJun20:
Remove duplicate names from given String Array, print names in given format with its length. Input Order has to be maintained.
String input[] = {"Aashtha","Kartikey","Piyush","Raj","Suparna","Piyush","Aashtha", "Barkha"}

Output : Aashtha -> 7
              Kartikey -> 8
              Piyush -> 6
	      Raj -> 3
	      Suparna -> 7
	      Barkha -> 6
Total unique elements are :- 6

Note: Use set interface. 
image


harshvegada (approved):
very good job @BarkhaPJun20 

Just siggestion from my side, if you use some existing methods when it comes to coding, like below just FYI

Set names = new LinkedHashSet(Arrays.asList(names));

If this looks complex then fine, no need to worry about this, your code is give very clear picture what you are doing. Good job @BarkhaPJun20

PR #1221, opened 2020-07-31
Aavruti_Assignment_36

AavrutiDJun20:
Remove duplicate names from given String Array, print names in given format with its length. Input Order has to be maintained. Find total Unique Numbers.

image


AditiGJUN20 (approved):
Code looks good to me.

PR #1223, opened 2020-07-31
Vaishnavi_Assignment_36_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20_Collection/RemoveDuplicates.java
+++ b/src/vaishnaviVJun20_Collection/RemoveDuplicates.java
@@ -1,0 +1,42 @@
+package vaishnaviVJun20_Collection;
+
+import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


harshvegada (approved):
Very well @VaishnaviVJun20 

you should use some existing methods when code to coding, like below just FYI

Set<String> names = new LinkedHashSet<String>(Arrays.asList(input));

If this looks complex then fine, no need to worry about this, your code is give very clear picture what you are doing. Good job

VaishnaviVJun20:
> Very well @VaishnaviVJun20
> 
> you should use some existing methods when code to coding, like below just FYI
> 
> Set names = new LinkedHashSet(Arrays.asList(input));
> 
> If this looks complex then fine, no need to worry about this, your code is give very clear picture what you are doing. Good job

Thank you @harshvegada I will definitely use this method next time.

PR #1224, opened 2020-07-31
Palak_Assignment36_20Min

PalakSJun20:
M not sure it is correct way to use Arrays.asList over here 
Kindly Provide comments
 
image


AbhishekSJune2020:
code looks good to me.

PR #1229, opened 2020-07-31
Archana_Assignment_36_Done

ArchanaNJun20:
Upadated output:
image

### DIFF
--- a/src/archanaNJun20/RemoveDuplicateNames.java
+++ b/src/archanaNJun20/RemoveDuplicateNames.java
@@ -1,19 +1,19 @@
 package archanaNJun20;
 
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
 public class RemoveDuplicateNames {
 	void removeDuplicateNamesFromString(String[] input) {
-		Set<String> nameString = new LinkedHashSet<>();
-		for(String name:input) 
-			nameString.add(name);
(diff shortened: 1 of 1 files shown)


src/archanaNJun20/RemoveDuplicateNames.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+public class RemoveDuplicateNames {
+	void removeDuplicateNamesFromString(String[] input) {
+		Set<String> nameString = new LinkedHashSet<>();
+		for(String name:input) 
+			nameString.add(name);
```
AavrutiDJun20: You can directly write like this instead of using loop
`Set<String> inputSet = new LinkedHashSet<String>(Arrays.asList(input));`

src/archanaNJun20/RemoveDuplicateNames.java
```diff
@@ -0,0 +1,22 @@
...
+		Iterator<String> iterator = nameString.iterator();
+		while (iterator.hasNext()) {
+			String temp=iterator.next();
+			System.out.println(temp+"->"+temp.length());
+		}
```
AavrutiDJun20, line 15: we have to find and print "Total unique elements are :- 6" as well

ArchanaNJun20: Code is updated based on your comment.Thank you

AavrutiDJun20 (approved):
Code looks good to me.

PR #1246, opened 2020-08-01
Pranita_Assignment_36_Done_15mins

PranitaPJun20:
image


harshvegada (approved):
Right approach @PranitaPJun20 
As per the expectation

PR #1251, opened 2020-08-02
Anup_Coding_Exam_Done_20mins

AnupSJun20:
image

### DIFF
--- a/src/anupSJun20/StringComparison.java
+++ b/src/anupSJun20/StringComparison.java
@@ -21,7 +21,7 @@ public class StringComparison {
 	void findRotationRequired(String str1, String str2) {
 		int count = 0;
 		while(!str1.equalsIgnoreCase(str2)) {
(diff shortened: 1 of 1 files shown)


src/anupSJun20/StringComparison.java
```diff
@@ -0,0 +1,37 @@
...
+efadbcd
+defadbc
+String input2 = "defadbc";
+output : 3*/
+
+public class StringComparison {
+	
+	void findRotationRequired(String str1, String str2) {
+		int count = 0;
+		while(!str1.equalsIgnoreCase(str2)) {
```
KrishnaKTechnocredits, line 23: For positive scenario it looks super good. Just a suggestion. 

add one more condition in while loop to handle negative scenario as well otherwise if two string could not compare at all then it would be infinite loop. 

while(!str1.equalsIgnoreCase(str2) && count <= input1.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1252, opened 2020-08-02
Nikhil_Coding_Exam_10_Done_15Min

NikhilMJun20:
-Completed and pushed Coding Exam # 10
-Please review and grade.

-PFB screenshot of output of the code:-

Annotation 2020-08-02 115602

### DIFF
--- a/src/nikhilMJun20/StringEquality.java
+++ b/src/nikhilMJun20/StringEquality.java
@@ -1,0 +1,38 @@
+/*
+Coding test 10 :
+
+Print minimum rotation required to compare both strings equality.
+
+Example 1 :
+String input1 = "abcdef";
+String input2 = "defabc";
+output : 3
+
+Example 2 :
+String input1 = "adbcdef"; -> push right
+fadbcde
+efadbcd
+defadbc
+String input2 = "defadbc";
+output : 3
+*/
+
+package nikhilMJun20;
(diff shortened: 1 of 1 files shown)


src/nikhilMJun20/StringEquality.java
```diff
@@ -0,0 +1,38 @@
...
+String input2 = "defadbc";
+output : 3
+*/
+
+package nikhilMJun20;
+
+public class StringEquality {
+	int getRotationCount(String input1, String input2) {
+		int count = 0;
+		while(! input1.equals(input2)) {
```
KrishnaKTechnocredits, line 25: @NikhilMJun20 , For positive scenario it looks super good. 
Just a suggestion.
we should add one more condition in while loop to handle negative scenario as well otherwise if two string could not compare at all then it would be infinite loop.

while(!input1.equals(input2) && count <= input1.length)

KrishnaKTechnocredits (approved):
Code looks good to me. Please take a look at my comments.

src/nikhilMJun20/StringEquality.java
NikhilMJun20: @KrishnaKTechnocredits 👍 
Yes, I missed this condition.
I will keep in mind...

PR #1254, opened 2020-08-02
Swapna_Coding_Exam_10_Done_25_min

SwapnaWJun20:
image

### DIFF
--- a/src/swapnaWJun20/RotationEqualString.java
+++ b/src/swapnaWJun20/RotationEqualString.java
@@ -1,0 +1,46 @@
+/*Coding test 10 :
+
+Print minimum rotation required to compare both strings equality.
+
+Example 1 :
+String input1 = "abcdef";
+String input2 = "defabc";
+output : 3
+
+Example 2 :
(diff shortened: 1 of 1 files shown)


src/swapnaWJun20/RotationEqualString.java
```diff
@@ -0,0 +1,46 @@
...
+		else {
+			while (index < input1.length()) {
+				input1 = input1.substring(input1.length() - 1) + input1.substring(0, input1.length() - 1);
+				System.out.println(input1 + " : at rotation " + index);
+				if (input1.equals(input2)) {
+					System.out.println("\nMinimum rotation required : " + (index + 1));
+					return;
+				}
+				index++;
+			}
```
KrishnaKTechnocredits, line 36: If you add one statement here after while loop, saying String cant be made identical, then it would handle negative scenarios too. Think about it.

KrishnaKTechnocredits (approved):
Nice , clean code. looks good to me. Please take a look at comments.

PR #1257, opened 2020-08-02
Anshu-Exam-Code-10-15min-Done

AnshuKJun20:
image

### DIFF
--- a/src/anshuKJun20/AppendLastString.java
+++ b/src/anshuKJun20/AppendLastString.java
@@ -1,0 +1,24 @@
+package anshuKJun20;
+
+public class AppendLastString {
+	
+	void appendLastString(String input ,String requiredOutput) {
+		int count =0;
+		if(input.length() == requiredOutput.length()) {
+			while(!input.equals(requiredOutput)) {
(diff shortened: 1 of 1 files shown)


src/anshuKJun20/AppendLastString.java
```diff
@@ -0,0 +1,24 @@
+package anshuKJun20;
+
+public class AppendLastString {
+	
+	void appendLastString(String input ,String requiredOutput) {
+		int count =0;
+		if(input.length() == requiredOutput.length()) {
```
KrishnaKTechnocredits, line 7: you should add else part too. if it is not equal tell the user it cant be made identical.

src/anshuKJun20/AppendLastString.java
```diff
@@ -0,0 +1,24 @@
...
+			while(!input.equals(requiredOutput)) {
```
KrishnaKTechnocredits, line 8: what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!input.equals(requiredOutput) && count< input.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1258, opened 2020-08-02
Abhishek_codingExam_10_25minutes

AbhishekSJune2020:
Two Strings Input are : 
abcdef
defabc
minimum rotation required : 3

### DIFF
--- a/src/abhishekSJun20/CodingExam10.java
+++ b/src/abhishekSJun20/CodingExam10.java
@@ -1,0 +1,34 @@
+package abhishekSJun20;
+
+/*
+ * Print minimum rotation required to compare both strings equality.
+
+Example 1  : 
+String input1 = "abcdef";
+String input2 = "defabc";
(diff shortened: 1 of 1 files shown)


src/abhishekSJun20/CodingExam10.java
```diff
@@ -0,0 +1,34 @@
...
+fadbcde
+efadbcd
+defadbc
+String input2 = "defadbc";
+output : 3
+ */
+public class CodingExam10 {
+	void findRotationRequired(String input1, String input2) {
+		int count = 0;
+		while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 21: what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!input1.equals(input2) && count< input.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1259, opened 2020-08-02
Vaishnavi_Coding_Exam_10_30min_Done

VaishnaviVJun20:
image

### DIFF
--- a/src/vaishnaviVJun20_Coding_Exams/Rotation.java
+++ b/src/vaishnaviVJun20_Coding_Exams/Rotation.java
@@ -1,0 +1,39 @@
+package vaishnaviVJun20_Coding_Exams;
+
+/* Coding test 10 :
+
+Print minimum rotation required to compare both strings equality.
+
+Example 1  : 
+String input1 = "abcdef";
+String input2 = "defabc";
+output : 3
+
+Example 2  : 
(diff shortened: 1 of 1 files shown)


src/vaishnaviVJun20_Coding_Exams/Rotation.java
```diff
@@ -0,0 +1,39 @@
...
+efadbcd
+defadbc
+String input2 = "defadbc";
+output : 3  */
+
+public class Rotation {
+	
+	void findMinimunRotation(String input1, String input2) {
+		int rotation=0;
+		 while(!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 24: what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!input1.equals(input2) && rotation< input.length)

KrishnaKTechnocredits (approved):
Code looks good to me. Please take a look at my comments

src/vaishnaviVJun20_Coding_Exams/Rotation.java
VaishnaviVJun20: Thank you ! @KrishnaKTechnocredits  
I will add this condition in my practice programs :)

PR #1260, opened 2020-08-02
Barkha_CodingExam_10_15mins

BarkhaPJun20:
Print minimum rotation required to compare both strings equality.

Example 1  : 
String input1 = "abcdef";
String input2 = "defabc";
output : 3

Example 2  : 
String input1 = "adbcdef"; -> push right
fadbcde
efadbcd
defadbc
String input2 = "defadbc";
output : 3
image

### DIFF
--- a/src/barkhaPJun20/CodingExam10_PrintMinRotation.java
+++ b/src/barkhaPJun20/CodingExam10_PrintMinRotation.java
@@ -1,0 +1,33 @@
+/*Print minimum rotation required to compare both strings equality
+Example 1  : 
+String input1 = "abcdef";
+String input2 = "defabc";
+output : 3
+
+Example 2  : 
+String input1 = "adbcdef"; -> push right
+fadbcde
+efadbcd
+defadbc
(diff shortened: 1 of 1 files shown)


src/barkhaPJun20/CodingExam10_PrintMinRotation.java
```diff
@@ -0,0 +1,33 @@
...
+String input2 = "defadbc";
+output : 3 */
+
+package barkhaPJun20;
+
+public class CodingExam10_PrintMinRotation {
+	int count = 0;
+
+	void printRotation(String name1, String name2) {
+		while (!name1.equals(name2)) {
```
KrishnaKTechnocredits, line 21: what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!input1.equals(input2) && count< input.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

KrishnaKTechnocredits (approved):

PR #1261, opened 2020-08-02
Rachana_Coding_Exam_10_Done_20min

RachanaGJun20:
image

### DIFF
--- a/src/rachanaGJun20/CoadingExam/CodingExam10.java
+++ b/src/rachanaGJun20/CoadingExam/CodingExam10.java
@@ -1,0 +1,40 @@
+/*Coding test 10 :
+
+Print minimum rotation required to compare both strings equality.
+
+Example 1  : 
+String input1 = "abcdef";
+String input2 = "defabc";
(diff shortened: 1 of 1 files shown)


src/rachanaGJun20/CoadingExam/CodingExam10.java
```diff
@@ -0,0 +1,40 @@
...
+package rachanaGJun20.CoadingExam;
+
+public class CodingExam10 {
+	
+	static void compareString() {
+		String input1 = "abcdef";
+		String input2 = "defabc";
+		int iterationCount = 0;
+		if (input1.length() == input2.length()) {
+			while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 26: what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!input1.equals(input2) && iterationCount< input.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1262, opened 2020-08-02
ADT_CodingExam_10_22Mins

AditiGJUN20:
Screen Shot 2020-08-01 at 11 42 05 PM

### DIFF
--- a/src/aditiGJUN20/StringRotation.java
+++ b/src/aditiGJUN20/StringRotation.java
@@ -18,7 +18,7 @@ output : 3*/
 package aditiGJUN20;
 
 public class StringRotation {
-	int isStringRotated(String input1, String input2) {
+	int getStringRotationCount(String input1, String input2) {
 		int count = 0;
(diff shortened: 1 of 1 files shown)


src/aditiGJUN20/StringRotation.java
```diff
@@ -0,0 +1,44 @@
...
+
+package aditiGJUN20;
+
+public class StringRotation {
+	int getStringRotationCount(String input1, String input2) {
+		int count = 0;
+		if (input1.equals(null) || input2.equals(null) || input1.length() != input2.length())
+			System.out.println("Strings are not matching with any rotation");
+		else {
+			while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 26: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

KrishnaKTechnocredits (approved):
Code looks good to me. Keep it up

PR #1263, opened 2020-08-02
Deavina_Codingexam_10_25mins

DeavinaSJun20:
image

### DIFF
--- a/src/deavinaSJun20/StringMatchingClass.java
+++ b/src/deavinaSJun20/StringMatchingClass.java
@@ -1,0 +1,32 @@
+package deavinaSJun20;
+
+/* Print minimum rotation required to compare both strings equality.
+
+
+Example 1 :
+String input1 = "abcdef";
+String input2 = "defabc";
+output : 3
+*/
+public class StringMatchingClass {
+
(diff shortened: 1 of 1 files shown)


src/deavinaSJun20/StringMatchingClass.java
```diff
@@ -0,0 +1,32 @@
...
+Example 1 :
+String input1 = "abcdef";
+String input2 = "defabc";
+output : 3
+*/
+public class StringMatchingClass {
+
+	int matchString(String input1, String input2) {
+		int rotation_count = 0;
+		while(! input1.equals(input2)) {
```
KrishnaKTechnocredits, line 15: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!input1.equals(input2) && rotation_count< input1.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

src/deavinaSJun20/StringMatchingClass.java
DeavinaSJun20: Thanks for suggestion. I will practice more and more.

PR #1264, opened 2020-08-02
Pranita_Coding_Exam_10_Done_40Mins

PranitaPJun20:
image

### DIFF
--- a/src/pranitaPJun20/Coding_Exam_10.java
+++ b/src/pranitaPJun20/Coding_Exam_10.java
@@ -1,0 +1,40 @@
+/*
+ * Coding test 10 :
+
+Print minimum rotation required to compare both strings equality.
+
+Example 1  : 
+String input1 = "abcdef";
+String input2 = "defabc";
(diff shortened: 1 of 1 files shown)


src/pranitaPJun20/Coding_Exam_10.java
```diff
@@ -0,0 +1,40 @@
...
+String input2 = "defadbc";
+output : 3
+ */
+package pranitaPJun20;
+
+public class Coding_Exam_10 {
+
+	void CompareStringsAndfindMinimunRotation(String input1, String input2) {
+		int rotationCount = 0;
+		while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 25: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!input1.equals(input2) && rotationcount< input1.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1265, opened 2020-08-02
Karan_CodeExam_10_45mins

KARAN1595:
String input1 = "adbcdef"; -> push right
fadbcde
efadbcd
defadbc
String input2 = "defadbc";
output : 3

### DIFF
--- a/src/kARAN1595/StringEquality.java
+++ b/src/kARAN1595/StringEquality.java
@@ -1,0 +1,24 @@
+package kARAN1595;
+
+public class StringEquality {
+	int getRotationCount(String input1, String input2) {
+		int count = 0;
+		if (input1.length() == input2.length()) {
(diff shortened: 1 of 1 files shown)


src/kARAN1595/StringEquality.java
```diff
@@ -0,0 +1,24 @@
+package kARAN1595;
+
+public class StringEquality {
+	int getRotationCount(String input1, String input2) {
+		int count = 0;
+		if (input1.length() == input2.length()) {
+			while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 7: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1266, opened 2020-08-02
Shruti coding exam 10-35 mints

ShrutiDJun20:


src/shrutiDJun20CodeExams/CodeExam10.java
```diff
@@ -0,0 +1,68 @@
...
+					+ count);
+		} else {
+			System.out.println("Inputs provided are of diff length");
+		}
+	}
+
+	/*void rotateElementWithoutCollection(String input1, String input2) {
+		int count = 0;
+		if (input1.length() == input2.length()) {
+			while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 42: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1267, opened 2020-08-02
Raj_CodingTest_10_Done_30mins

RajAJun20:
CodingTest_10

### DIFF
--- a/src/rajAJun20/CodingTest10.java
+++ b/src/rajAJun20/CodingTest10.java
@@ -1,0 +1,33 @@
+/*Print minimum rotation required to compare both strings equality.
+Example 2  : 
+String input1 = "adbcdef"; -> push right
+fadbcde
+efadbcd
+defadbc
+String input2 = "defadbc";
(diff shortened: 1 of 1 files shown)


src/rajAJun20/CodingTest10.java
```diff
@@ -0,0 +1,33 @@
...
+
+package rajAJun20;
+
+public class CodingTest10 {
+	void findNumbersOfRotation(String str1,String str2) {
+		int count = 0;
+		if(str1.length()!=str2.length()){
+			System.out.println("Given strings are not identical..!!");
+		}else {	
+			while(!str1.equals(str2)) {
```
KrishnaKTechnocredits, line 18: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!str1.equals(str2) && count< str1.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1268, opened 2020-08-02
Madhura_Coding_Exam_10_40mins

MadhuraMJun20:
CodingExam10

### DIFF
--- a/src/madhuraMJun20/assignment30Plus/Coding_Exam_10.java
+++ b/src/madhuraMJun20/assignment30Plus/Coding_Exam_10.java
@@ -1,0 +1,34 @@
+/*Print minimum rotation required to compare both strings equality.
+Example 1  : 
+String input1 = "abcdef";
+String input2 = "defabc";
+output : 3
+
+Example 2  : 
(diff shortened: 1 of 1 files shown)


src/madhuraMJun20/assignment30Plus/Coding_Exam_10.java
```diff
@@ -0,0 +1,35 @@
...
+defadbc
+String input2 = "defadbc";
+output : 3*/
+package madhuraMJun20.assignment30Plus;
+
+public class Coding_Exam_10 {
+	static int getMinimumRotation(String input1, String input2) {
+		int rotationCount = 0;
+		if (input1.length() == input2.length()) {
+			while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 20: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

while(!input1.equals(input2) && rotationCount< input1.length)

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1269, opened 2020-08-02
Kartikey Coding test 10 25-30 mins

KartikeyDJune20:
image

### DIFF
--- a/src/kartikeyDJune20/MinimumRotationTest.java
+++ b/src/kartikeyDJune20/MinimumRotationTest.java
@@ -1,0 +1,23 @@
+package kartikeyDJune20;
+
+public class MinimumRotationTest {
+	public static void main(String[] args) {
+		String str1= "abcdef";
+		String str2= "defabc";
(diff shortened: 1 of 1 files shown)


src/kartikeyDJune20/MinimumRotationTest.java
```diff
@@ -0,0 +1,23 @@
...
+		int count=0;
+		for(int index=0;index<str1.length();index++) {
+			if(!str1.equals(str2)) {
+			str1=str1.charAt(str1.length()-1)+str1.substring(0,str1.length()-1);
+			count++;
+			}
+			else
+				break;			
+		}
+		System.out.println(count +": no. of roation is required to match");
```
KrishnaKTechnocredits, line 16: what if String didnt match at all. it will say 0 count is required to match a string. Think about it, how can you handle it.

KrishnaKTechnocredits (approved):
Code looks good to me, please take a look at my comments.

PR #1270, opened 2020-08-02
Palak_Coding_Exam10_40Min

PalakSJun20:
image

### DIFF
--- a/src/palakSJun20/StringRotationCheck.java
+++ b/src/palakSJun20/StringRotationCheck.java
@@ -1,0 +1,34 @@
+/*Coding test 10 :
+Print minimum rotation required to compare both strings equality.
+Example 2 :
+String input1 = "adbcdef"; -> push right
+fadbcde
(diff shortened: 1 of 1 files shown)


src/palakSJun20/StringRotationCheck.java
```diff
@@ -0,0 +1,34 @@
...
+package palakSJun20;
+
+public class StringRotationCheck {
+
+	public void findMinimumRotattion(String input1, String input2) {
+		int count = 0;
+		if (input1.length() != input2.length())
+			System.out.println("String lengths are not same");
+		else {
+			while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 18: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1271, opened 2020-08-02
Archana_Coding_Exam_Done_40mins

ArchanaNJun20:
image

### DIFF
--- a/src/archanaNJun20/CodingExam10.java
+++ b/src/archanaNJun20/CodingExam10.java
@@ -1,0 +1,18 @@
+package archanaNJun20;
+
+public class CodingExam10 {
+	void printRotation(String input1, String input2) {
+		int count = 0;
+		while (!input1.equals(input2)) {
+			char last = input1.charAt(input1.length() - 1);
+			input1 = last + input1.substring(0, (input1.length() - 1));
+			count++;
+		}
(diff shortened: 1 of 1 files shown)


src/archanaNJun20/CodingExam10.java
```diff
@@ -0,0 +1,18 @@
+package archanaNJun20;
+
+public class CodingExam10 {
+	void printRotation(String input1, String input2) {
+		int count = 0;
+		while (!input1.equals(input2)) {
+			char last = input1.charAt(input1.length() - 1);
+			input1 = last + input1.substring(0, (input1.length() - 1));
```
KrishnaKTechnocredits, line 8: we can merge line 7 and 8.

src/archanaNJun20/CodingExam10.java
```diff
@@ -0,0 +1,18 @@
+package archanaNJun20;
+
+public class CodingExam10 {
+	void printRotation(String input1, String input2) {
+		int count = 0;
+		while (!input1.equals(input2)) {
```
KrishnaKTechnocredits, line 6: Just a suggestion. what if String can't be made equals even after all possible rotation ?? this will run infinitely ...

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1272, opened 2020-08-02
Nikhil_Assignment_37_Done_30Min

NikhilMJun20:
-Completed and pushd Assignment # 37.
-Please review and comment.

-PFB screenshot of output of the program:-

Annotation 2020-08-02 211414

### DIFF
--- a/src/nikhilMJun20/StringFrequency.java
+++ b/src/nikhilMJun20/StringFrequency.java
@@ -1,0 +1,49 @@
+/*
+Assignment-37 : 2nd Aug'2020
+
+Input : Hi Globant techno Hi Hello Globant techno techno Globant techno
+FR - 1 : freq of each word
+FR - 2 : Maxmimum repeated word and its frequency
+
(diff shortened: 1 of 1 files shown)


src/nikhilMJun20/StringFrequency.java
```diff
@@ -0,0 +1,49 @@
...
+
+package nikhilMJun20;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+
+public class StringFrequency {
+	void findFrequency(String input) {
+		ArrayList<String> strList = new ArrayList<String> (Arrays.asList(input.split(" ")));
```
VaishnaviVJun20, line 23: --logic is correct
--However here you can directly use array as well String[] array = str.split(" ");
--Nothing wrong in using ArrayList as well.

VaishnaviVJun20 (approved):
--Code looks good

PR #1275, opened 2020-08-03
AMITA_Assignment_37_25mins

AmitaRJun20:
Please Review Assignment 37...
MaxCountStringFreqHashMap


MaheshKJun20 (approved):
Hi Amita,

Requirement was to get count of words from Statement(String). so need to add 1 step convert string in array. Rest all logic is good.

PR #1286, opened 2020-08-03
Anup_Assignment_37_Done_20mins_PR

AnupSJun20:
Please review
image


harshvegada (approved):
Correct @AnupSJun20 
very well, keep doing well,
Output and coding both are as per the expectation

PR #1288, opened 2020-08-04
Pranita_Assignment_37_Done_30mins

PranitaPJun20:
image

### DIFF
--- a/src/pranitaPJun20/MaxFreqWordString_Collections.java
+++ b/src/pranitaPJun20/MaxFreqWordString_Collections.java
@@ -12,7 +12,6 @@ Maximum repeated word is techno having frequency 4.
 
 package pranitaPJun20;
 
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


src/pranitaPJun20/MaxFreqWordString_Collections.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class MaxFreqWordString_Collections {
+
+	void findMaxRepeatedWordFrequency(String input) {
+		String[] inputArray = input.split(" ");
+		int count = 0;
+		Map<String, Integer> wordMap = new LinkedHashMap<String, Integer>();
+		for (String name : inputArray) {
+			if (wordMap.containsKey(name)) {
+				wordMap.put(name, wordMap.get(name) + 1);
```
harshvegada, line 28: Hi @PranitaPJun20 
Here you need to write

Line 39 to 42

What you write that is also option but we discuss Try to check with in if condition.

harshvegada (changes requested):
Please do change as per comment

src/pranitaPJun20/MaxFreqWordString_Collections.java
PranitaPJun20: Ok Harsh.Will update my code

harshvegada (approved):
Looks good pranita
Keep it up 👍

PR #1295, opened 2020-08-31
Aavruti_Coding_Exam_ValidatePIN

AavrutiDJun20:
image

### DIFF
--- a/src/aavrutiDJun20/ValidatePinNumber.java
+++ b/src/aavrutiDJun20/ValidatePinNumber.java
@@ -1,0 +1,62 @@
+package aavrutiDJun20;
+
+import java.util.Arrays;
+
+public class ValidatePinNumber {
+	
+	void validatePin(String number) {
+		String num = "";
+		boolean flag = true;
+		
+		try {
+			Integer.parseInt(number);
+			
+			for(int index=0;index<number.length();index++) {
+				if(num.contains(String.valueOf(number.charAt(index)))) {
+					flag = false;
+					break;
+				}
+				else {
+					num += number.charAt(index);
+				}
+			}
+			if(flag) {
+				char charArray[] = number.toCharArray();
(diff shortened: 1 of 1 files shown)


src/aavrutiDJun20/ValidatePinNumber.java
```diff
@@ -0,0 +1,62 @@
...
+					num += number.charAt(index);
+				}
+			}
+			if(flag) {
+				char charArray[] = number.toCharArray();
+				Arrays.sort(charArray);
+				String numSorted = String.valueOf(charArray);
+				
+				StringBuilder strReverse = new StringBuilder(numSorted);
+				strReverse.reverse();
```
harshvegada, line 29: Hi @AavrutiDJun20 
I feel using this reverse() method we not achieving our goal, Here we need to check ascending and descending order, not reverse order. check once

harshvegada (approved):
very nice @AavrutiDJun20 
As expected

src/aavrutiDJun20/ValidatePinNumber.java
```diff
@@ -0,0 +1,62 @@
...
+	
+	void validatePin(String number) {
+		String num = "";
+		boolean flag = true;
+		
+		try {
+			Integer.parseInt(number);
+			
+			for(int index=0;index<number.length();index++) {
+				if(num.contains(String.valueOf(number.charAt(index)))) {
```
KrishnaKTechnocredits, line 15: Instead of writing loop you can use try catch --> NumberFormateException

KrishnaKTechnocredits: try{
flag= true
}catch(NumberFormateException ne){
flag = false
}

KrishnaKTechnocredits (approved):
Code looks good to me.

PR #1296, opened 2020-08-31
Mahesh_Coding_Exam_31_Aug_40Mins

MaheshKJun20:
image

### DIFF
--- a/src/maheshKJun20/ValidateAtmPin.java
+++ b/src/maheshKJun20/ValidateAtmPin.java
@@ -1,0 +1,79 @@
+package maheshKJun20;
+
+public class ValidateAtmPin {
+	
+	void validatePin(String pin) {
+		if (validateLength(pin)) {
+			if (validateDigit(pin)) {
+				if (validateNumRepitition(pin))
+					if (validateAscendingOrder(pin) || validateDescendingOrder(pin))
+						System.out.println("ATM pin " + pin + " is Valid");
+					else
+						System.out.println("Invalid pin. Order Should be Ascending or Desending");
+			}				
+		}		
+	}
+	
+	boolean validateLength(String pin) {
+		if (pin.length() == 4)
+			return true;
+		else {
+			System.out.println("Invalid pin. Length of pin: " + pin.length());
+			return false;
+		}
+	}
+	
+	boolean validateDigit(String pin) {
+		for (int index=0; index<pin.length();index++) {
+			if(!Character.isDigit(pin.charAt(index))){
+				System.out.println("Invalid pin. Invalid character in the pin '" + pin.charAt(index) + "'");
+				return false;
+			}
+		}
+		return true;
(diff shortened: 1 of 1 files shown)


src/maheshKJun20/ValidateAtmPin.java
```diff
@@ -0,0 +1,79 @@
...
+		if (pin.length() == 4)
+			return true;
+		else {
+			System.out.println("Invalid pin. Length of pin: " + pin.length());
+			return false;
+		}
+	}
+	
+	boolean validateDigit(String pin) {
+		for (int index=0; index<pin.length();index++) {
```
KrishnaKTechnocredits, line 27: Instead of writing loop you can use try catch --> NumberFormateException
Think about it.

src/maheshKJun20/ValidateAtmPin.java
```diff
@@ -0,0 +1,79 @@
...
+			if(!Character.isDigit(pin.charAt(index))){
+				System.out.println("Invalid pin. Invalid character in the pin '" + pin.charAt(index) + "'");
+				return false;
+			}
+		}
+		return true;
+	}
+	
+	boolean validateAscendingOrder(String pin) {
+		for (int index=0; index<pin.length();index++) {
```
KrishnaKTechnocredits, line 37: within single loop we can do it. why do we compare all numbers.
just compare first to second index, second to third and so on.. think about it .. you dont need to compare all 1 number with all previous numbers.. share your thoughts.

src/maheshKJun20/ValidateAtmPin.java
```diff
@@ -0,0 +1,79 @@
...
+					return false;
+				}
+			}
+		}
+		return true;
+	}
+	
+	boolean validateNumRepitition(String pin) {
+		for (int index=0; index<pin.length();index++) {
+			for (int innerIndex=index+1; innerIndex<pin.length();innerIndex++) {
```
harshvegada, line 62: why do we need two for loop here?
we can do using one right

use indexOf & lastIndexOf

harshvegada (approved):
pls check once

src/maheshKJun20/ValidateAtmPin.java
MaheshKJun20: @harshvegada  - Yes, one loop with indexOf and lastIndexOf method would have been a better solution to this. Noted and will amend the changes. Thanks for the feeback

src/maheshKJun20/ValidateAtmPin.java
MaheshKJun20: @KrishnaKTechnocredits  - Yes Krishna, That would have been more appropriate solution for this. Will update the changes. 
For NumberFormatException - Realized it after submition. Will update that as well in the code.
Thanks for the Feedback.

PR #1306, opened 2020-08-31
Barkha_JavaTest_13 Aug

BarkhaPJun20:
image

### DIFF
--- a/src/barkhaPJun20/PinValidation.java
+++ b/src/barkhaPJun20/PinValidation.java
@@ -1,0 +1,74 @@
+/*WAP to validate Pin number.
+
+Pin number is valid if it follows below rules :
+1) ATM pin length should be 4.
+2) All characters must be digit.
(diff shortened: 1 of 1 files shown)


harshvegada (commented):
what about other options

src/barkhaPJun20/PinValidation.java
```diff
@@ -0,0 +1,74 @@
...
+It would be great if you think solution by below approach :
+-> One with collections.
+-> Without using collections. */
+package barkhaPJun20;
+
+public class PinValidation {
+
+	boolean isDigit(String pin) {
+		for (int index = 0; index < pin.length(); index++) {
+			if (Character.isDigit(pin.charAt(index)))
```
harshvegada, line 26: you should use isLetter() instead of isDigit and second point return false if any char is letter

harshvegada (commented):
other method as expected, please add more number

PR #1313, opened 2020-09-24
Barkha_CodingExam

BarkhaPJun20:
image

### DIFF
--- a/src/barkhaPJun20/HighestTripletSum.java
+++ b/src/barkhaPJun20/HighestTripletSum.java
@@ -6,6 +6,7 @@ package barkhaPJun20;
 public class HighestTripletSum {
 	int highestSum = 0;
 	int sum;
+	String triplets;
 
 	public void SumOfTriplet(int[] num) {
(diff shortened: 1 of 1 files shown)


src/barkhaPJun20/HighestTripletSum.java
```diff
@@ -0,0 +1,29 @@
...
+
+	public void SumOfTriplet(int[] num) {
+		for (int index = 0; index < num.length; index++) {
+			sum = num[index] + num[index + 1] + num[index + 2];
+
+			if (sum > highestSum) {
+				highestSum = sum;
+				triplets=num[index]+", " + num[index + 1]+", " + num[index + 2];
+			}
+			index += 2;
```
harshvegada, line 19: Hi @BarkhaPJun20 
Excellent work,
Improvement part : Instead of increase value on line 19th you can add this in loop line 13th (instead of i++ you can write index+=3).

harshvegada (approved):
Keep doing well

